PR #38, opened 2014-08-19
Logging improvements

sgryphon:
Error logging was pretty rough, e.g. hard coded to write to the console (even if, for example, running in a web application).

I have changed it to use the standard .NET Framework System.Diagnostics with a named TraceSource "OfficeDevPnP.Core".

Tracing should generally be an internal system, however it was public so I have kept the old behaviour (marked Obsolete), including automatically configuring writing to the console if needed. This means that all existing samples should continue to operate the same (with slightly different output format). (It also probably shouldn't be a partial class, but I have left that for now as well).

Using TraceSource means the level of diagnostics can be configured and directed to the console, ASP.NET trace, or other listeners (including third part like log4net or NLog). These changes will give a good base for moving forward, adding things like activity scopes and correlation IDs.

### DIFF
--- a/OfficeDevPnP.Core/OfficeDevPnP.Core/Utilities/LoggingUtility.cs
+++ b/OfficeDevPnP.Core/OfficeDevPnP.Core/Utilities/LoggingUtility.cs
@@ -111,7 +111,7 @@ namespace OfficeDevPnP.Core.Utilities
         /// </summary>
         /// <param name="message">Message to be logged</param>
         /// <param name="category">Category to be used for the logged message</param>
-        //[Obsolete("Create an instance with a named trace source and use TraceVerbose()")]
+        [Obsolete("Create a named instance or, internally, use LoggingUtility.Internal.TraceVerbose()")]
         [EditorBrowsable(System.ComponentModel.EditorBrowsableState.Never)]
         public static void LogVerbose(string message, EventCategory category) 
         {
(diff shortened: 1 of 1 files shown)


seriewe:
Hello,
can you please share some Config examples how I can Route the PnP loggs to NLog?

I'm able to log the Output of my own logger and the PnP logs to a File, but then I see just the PnP logs inside the Debug output window. As soon as I add a Console target to the NLog config, I get a StackOverflow Exception.

This is my web.config:
 <sectionGroup name="pnp">
      <section name="tracing" type="OfficeDevPnP.Core.Diagnostics.LogConfigurationTracingSection, OfficeDevPnP.Core" 
               allowLocation="true" 
               allowDefinition="Everywhere" />
    </sectionGroup>
  
  <pnp>
    <tracing logLevel="Debug">
      <logger type="OfficeDevPnP.Core.Diagnostics.TraceLogger, OfficeDevPnP.Core, Culture=neutral, PublicKeyToken=null" />
    </tracing>
  </pnp>
  <system.diagnostics>
        <sharedListeners>
            <add name="NLogTraceListener" type="NLog.NLogTraceListener, NLog" />
            
        </sharedListeners>
        <sources>
            <source name="OfficeDevPnP.Core" switchValue="All">
                <listeners>
                    <add name="NLogTraceListener" type="NLog.NLogTraceListener, NLog" />
                </listeners>
            </source>
          </sources>
        
    </system.diagnostics>    
  

And a part of my NLog.config:
 <logger name="*" minlevel="Debug" writeTo="FileLogs" />
  <logger name="*" minlevel="Debug" writeTo="console" />

I did different combinations of listener source and so one, but I don't figure out how to log to the File and also have the output inside the output window inside VS.

Thanks ;)

PR #348, opened 2015-01-02
Group and context

erwinvanhunen:
Increased performance of cmdlets inheriting from SPOWebCmdlet
Introduced new Set-SPOGroup cmdlet
Deprecated -SetAssociatedGroup parameter on new-spogroup cmdlet. Moved the parameter to set-spogroup
Caches the initial context to restore after a SPOWebCmdlet with a -Web parameter is done executing.


jansenbe:
Thanks Erwin

PR #349, opened 2015-01-03
Added SharePoint Hosted App which will show how to consume the Office Graph API

vman:


jansenbe:
Thanks for this nice contribution!

jansenbe:
Merged. Did however change the way how images are loaded as the provided approach didn't work for me. Somehow userphoto.aspx in combination with accountname parameter and app web url doesn't work. The same combination with a host web url works fine. Explicitly providing the url to userphoto.aspx  seems to work in all cases.

vman:
Hmm that is interesting as it was working for me. Will have to look into this a bit more. Thanks a lot of fixing it bert!

PR #350, opened 2015-01-03
List cmdlets

erwinvanhunen:
Added new cmdlets to set list properties, break role inheritance and set list permissions.
Updated some inner workings of group cmdlets.


jansenbe:
Thanks Erwin

PR #354, opened 2015-01-05
added AllowAllPageLayouts method (similar to SSOM)

vman:
Added AllowAllPageLayouts method (similar to SSOM) which will allow all the page layouts to be used on the web. The ClearAvailablePageLayouts method can be used to achieve the same functionality but then the name of the method does not seem meaningful


jansenbe:
Good idea. I've merged this PR but made the old method obsolete and moved it to the .deprecated cs file

vman:
Thanks. I was going to mark the ClearAvailablePageLayouts method as deprecated but thought it also makes sense as a method on its own.

I am also further modifying the AllowAllPageLayouts method to accept an optional parameter which will push the changes to subwebs. Will commit it soon.

jansenbe:
good, ensure you refresh your branch with the update I just did.

jansenbe:
We're trying to streamline the API and make it easier to use, so having 2 methods for the same function are not needed. AllowAllPageLayouts in in similar to SSOM and makes more sense.

PR #359, opened 2015-01-07
Office365Api.Demo Updated to latest release of libraries

PaoloPia:
Guys,
I've update the sample to the latest libraries (Office 365 API .NET Client Library, ADAL and AD GraphClient).
However, there are some issues that we need to share and fix/decide how to handle.

Bert,
Please take a look at my comments in the code. You can find them by searching the "Note from @PaoloPia" trailer.

So far, the issues are:
1) MailApiSample: I can't find any method to get mail stats, as like as it was in the beta version of the libraries. Maybe it is my fault, maybe the PG removed that capability. Can you check it against the PG? I can't find anything anymore in the online docs on MSDN/Office Dev about mail stats.
2) MailApiSample: I removed the client-side paging, because now the server-side paging works properly. Do you agree?
3) SitesApiSample: the Site Collection files query doesn't work. I fighted against it more than two hours (!) ... but no win! :( Can you try to have a look at the failing call, which is highlighted with an IMPORTANT comment, and which is commented out of the MainWindow.xaml.cs code. Thanks.
4) I placed some query limits for emails and files, to not overload the Office 365 servers and the network in case of a huge number of items to download. I hope you agree.

After committing this demo, I'm planning to move to an ASP.NET MVC demo, as well. I think it would be really useful to have both the scenarios (native/desktop app versus web application). Can I? Do you agree?

Thanks,
Paolo

### DIFF
--- a/Samples/Office365Api.Overview/Office365Api.Demo/ActiveDirectoryApiSample.cs
+++ b/Samples/Office365Api.Overview/Office365Api.Demo/ActiveDirectoryApiSample.cs
@@ -1,5 +1,6 @@
 ﻿using Microsoft.Office365.OAuth;
-using Microsoft.Office365.ActiveDirectory;
+using Microsoft.Azure.ActiveDirectory.GraphClient;
+using Microsoft.IdentityModel.Clients.ActiveDirectory;
 using System;
 using System.Collections.Generic;
 using System.Linq;
@@ -10,21 +11,9 @@ namespace Office365Api.Demo
 {
     public static class ActiveDirectoryApiSample
     {
-        const string ServiceResourceId = "https://graph.windows.net/";
-        static readonly Uri ServiceEndpointUri = new Uri("https://graph.windows.net/");
-
-        // Do not make static in Web apps; store it in session or in a cookie instead
-        static string _lastLoggedInUser;
-        //static DiscoveryContext _discoveryContext;
-        public static DiscoveryContext _discoveryContext
-        {
-            get;
-            set;
-        }
-
         public static async Task<IEnumerable<IUser>> GetUsers()
         {
-            var client = await EnsureClientCreated();
+            var client = EnsureClientCreated();
 
             var userResults = await client.DirectoryObjects.OfType<User>().ExecuteAsync();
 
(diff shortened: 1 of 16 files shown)


PaoloPia:
I forgot it: of course you should remove my default values before committing the code. Sorry for not removing them.

jansenbe:
Hi Paolo,

I've been doing some investigation:
"1) MailApiSample: I can't find any method to get mail stats, as like as it was in the beta version of the libraries" **==> seems to be gone, but will seek confirmation**
"2) MailApiSample: I removed the client-side paging, because now the server-side paging works properly. Do you agree?" **==> Yep**
"3) SitesApiSample: the Site Collection files query doesn't work. I fighted against it more than two hours (!) ... but no win! :( Can you try to have a look at the failing call, which is highlighted with an IMPORTANT comment, and which is commented out of the MainWindow.xaml.cs code. Thanks. **==> I want to verify is it's still intended to use the O365 to read other SharePoint files besides the OneDrive files. Can't seem to find any recent samples for this**
 4) I placed some query limits for emails and files, to not overload the Office 365 servers and the network in case of a huge number of items to download. I hope you agree."  **==> sure**

After committing this demo, I'm planning to move to an ASP.NET MVC demo, as well. I think it would be really useful to have both the scenarios (native/desktop app versus web application). Can I? Do you agree? **==>Absolutely**

PaoloPia:
Hi Bert,
Thanks for your answers. Sorry for re-opening this thread, but I'm working on the MVC application and about question 3) I think that the SharePointClient should work against SPO, as well. In fact the "Connected Services" wizard allows configuring explicit permissions for SharePoint (Sites group), like "Read items in all site collections", "Have full control of all site collections", etc.
Moreover, I tried to trace the network traffic using Fiddler2 and, as far as I can see, the HTTP request is ok, as well as the JSON response. Nevertheless, the OData client library fails to deserialize the JSON response, trying to create a new instance of an abstract class (probably Microsoft.Office365.SharePoint.FileServices.Item).

I will try to investigate further on it. In the meantime, what about removing from the dev branch the failing code? I can clean-up the solution while creating the MVC app, if you like and agree.

Paolo

jansenbe:
Hi Paolo,

I've did some checking and the Sites API should still be possible but I've been told that it's limited to reading from the site's "Shared Documents" library only...Did some quick testing with the modified code, but so far no luck...will need to deep digger, but will be for later

PaoloPia:
Thanks,
I will continue investigating the issue on my side, while working on the MVC web app. BTW, sorry about that but ... being able to query only the "Shared Documents" is totally useless in my opinion :-) ... 

Thanks,
Paolo

jansenbe:
Yep, I completely agree. Will see if I can dig up something more.

PR #372, opened 2015-01-10
Added method GetListItems

nanddeepn:
Added method GetListItems to get the list items from a list by matching caml query.


jansenbe:
Hi Nanddeep. Thanks for this pull request. After careful deliberation in the core team we've decided to not merge this PR. The code as such is fine and does work but we want to avoid growing the PnP Core library with methods that can be easily done by using just a view lines of code. If you want to keep this extension method I suggest you add it to your personal set of extensions.

Hope you understand.

nanddeepn:
Hello Bert,
Thank you for the detailed explanation. :)

PR #373, opened 2015-01-10
Added method SetListPermissions

nanddeepn:
Added method SetListPermissions to give additional permissions to user on list.

### DIFF
--- a/OfficeDevPnP.Core/OfficeDevPnP.Core/AppModelExtensions/ListExtensions.cs
+++ b/OfficeDevPnP.Core/OfficeDevPnP.Core/AppModelExtensions/ListExtensions.cs
@@ -1,5 +1,6 @@
 ﻿using Microsoft.SharePoint.Client;
 using Microsoft.SharePoint.Client.Taxonomy;
+using Microsoft.SharePoint.Client.Utilities;
 using OfficeDevPnP.Core;
 using OfficeDevPnP.Core.Entities;
 using OfficeDevPnP.Core.Utilities;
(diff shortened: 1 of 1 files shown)


nanddeepn:
Hello Erwin,
Thank you for the suggestion. The hard-coded English text "Everyone" is passed as argument to method only in case of giving permission to all users.

I am thinking of any other alternative way to make the method a language agnostic. Please suggest, if you have any thoughts on it.

erwinvanhunen:
The best way I would say is not take a string/loginname as input of the method, but take a User object instead. That way there's no need to handle special cases such as 'Everyone'.

erwinvanhunen:
Even better would be to take a 'Principal' object instead of a User. Then it doesn't matter if you send in a group or a user and you can also drop the PrincipalType parameter.

nanddeepn:
Hello Erwin,
Thank you for the quick and good suggestion. :+1: 
I will make the changes and create a new pull request.

jansenbe:
What do you guys think about adding an overload for SetListPermissions that instead of loginname takes our enum BuiltInIdentity (see AddReaderAccessImplementation in securityExtensions). This overloaded method can then resolve either the "Everyone" or the "Everyone except external users" to the correct claim. Once we've the claim we can then call the original SetListPermissions. Ideally we update the SecurityExtensions AddReaderAccessImplementation method to isolate the BuiltInIdentity to claim code so that it can be used in both methods.

Additionally we could also foresee an overload that takes a string roleDefinitionName instead of RoleType --> this way we're identical to the current AddPermissionLevelTo(group/user) methods in SecurityExtensions.

What do you guys think?

nanddeepn:
Hello Bert,
Thank you for the suggestion. We can have this helpful overload method.

nanddeepn:
The new pull request with above changes is created here: https://github.com/OfficeDev/PnP/pull/380

PR #383, opened 2015-01-14
Changed reference for AppForSharePointWebToolkit

beersky:
According to guidance there shouldn't be a reference in packages to AppForSharePoint16WebToolkit instead AppForSharePointWebToolkit.

Read more here: https://github.com/OfficeDev/PnP/blob/master/Guidance/ConfigSpPHAppsForDistro.md


jansenbe:
Thanks. These AppForSharePoint16WebToolkit package references keep coming back, nice catch :)

PR #391, opened 2015-01-17
Fix alignment of dialog and overlay in taxonomypicker component/sample

pbjorklund:
When opening dialog and overlay on a page with more than 1 screen of content
the artefakts did not align. Replaced absolute positioning with fixed.

Tested on IE11 and Firefox 36.0a2


erwinvanhunen:
Thanks Patrik!

PR #399, opened 2015-01-20
Get the language agnostic pages library name

nanddeepn:
Currently, we get pages library as hard-coded EN value "Pages", which works only in EN sites.
Modified the code to get the pages library name in a language agnostic way.


nanddeepn:
Sorry, Some issue with code. I will fix it and recreate pull request.

PR #401, opened 2015-01-20
Taxonomy import fixes

erwinvanhunen:
Refactored ImportTerms method that supports the TermGroup|TermSet|Term format to utilize the ImportTerms functionality that supports the 'classic' CSV format. Result is easier to maintain and only one import routine instead of 2.
Removed 'force' parameter from the Import-SPOTaxonomy cmdlet and introduced an SynchronizeDeletions parameter, in sync with the ImportTerms method.


jansenbe:
Nice :)

PR #402, opened 2015-01-20
Added extension methods for publishing pages

nanddeepn:
Added the extension methods for publishing pages:
1. CheckOutPublishingPage
2. CheckInPublishingPage
3. AddWebPartsToPublishingPage


erwinvanhunen:
Hi Nanddeep,

We discussed your pull request in the core team and came to the conclusion that we will not merge the methods.

Your proposed Checkout and Checkin methods overlap or partly duplicate other methods already in place (for instance FileFolderExtensions.CheckoutFile and CheckinFile) and we feel it would only create more confusion adding these specific methods. 
Besides that, your proposed CheckInPublishingPage has many parameters. We think if you would propose a future method with that many parameters it's better to send in an entity instead.

The AddWebPartsToPublishingPage duplicates existing methods (AddWebPartToWeb/WikiPage) and that functionality can be easily done using the existing extension methods.

We nevertheless thank you for your efforts and look forward to other pull requests from you!

I will close this PR now.

nanddeepn:
Hello Erwin,
Thank you for the detailed explanation.

PR #410, opened 2015-01-28
Load property ContentTypesEnabled before calling method on it

nanddeepn:
Currently, the methods called separately throws exception: 
Field or property ContentTypesEnabled is not initialized.

### DIFF
--- a/OfficeDevPnP.Core/OfficeDevPnP.Core/AppModelExtensions/FieldAndContentTypeExtensions.cs
+++ b/OfficeDevPnP.Core/OfficeDevPnP.Core/AppModelExtensions/FieldAndContentTypeExtensions.cs
@@ -656,8 +656,11 @@ namespace Microsoft.SharePoint.Client
                 return;
             }
 
-            list.Context.Load(list, l => l.ContentTypesEnabled);
-            list.Context.ExecuteQuery();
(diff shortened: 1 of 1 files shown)


OfficeDevPnP.Core/OfficeDevPnP.Core/AppModelExtensions/FieldAndContentTypeExtensions.cs
```diff
@@ -592,10 +647,17 @@ public static void AddContentTypeToListByName(this List list, string contentType
...
+            {
                 throw new ArgumentNullException("contentType");
+            }
 
             if (list.ContentTypeExistsById(contentType.Id.StringValue))
+            {
                 return;
+            }
+
+            list.Context.Load(list, l => l.ContentTypesEnabled);
```
erwinvanhunen: What about check if the property is available first before loading it instead of assuming it's not and always loading it? E.g. 
```
if(!list.IsPropertyAvailable("ContentTypesEnabled") 
{ 
// load the property 
}
```

nanddeepn:
Hello Erwin,
Thank you for the valuable suggestion.
I will add the check to see IsPropertyAvailable, before loading it.

Thank you.

nanddeepn:
Implemented the new changes here: https://github.com/OfficeDev/PnP/pull/411
Closing this pull request

PR #431, opened 2015-02-06
Added support for configuration Search tab navigation on a search page

wobba:
Added support for doing search navigation using the navigation extensions which previously supported quicklaunch and top level. I switched the bool to enum, and kept old overloads - marked as obsolete.

Also did some code restructure to simplify the logic and nesting.

Added unit tests for all three navigation scenarios.

-Mikael Svenson (Puzzlepart)


jansenbe:
Thanks for your contribution Mikael, really appreciated! 

Will be looking into this right now.

wobba:
Thank you :) 

-----Original Message-----

PR #433, opened 2015-02-08
Rely on ?? operator at call site instead of passing default values to Get-methods

ronnieholm:
Removed the default argument to ListExtensions.GetPropertyBagValueString and ListExtensions.GetPropertyBagValueInt.

The default argument to GetPropertyBagValueString was never implemented and OfficeDevPnP.Core code calling GetPropertyBagValueString didn't rely on the default value behavior being implemented.

Instead of passing default values to GetPropertyBagValueString and GetPropertyBagValueString, why not use the build-in C# ?? operator at the call site?

For instance:

            int? x = 0;
            int? y = null;

            System.Console.WriteLine(x.HasValue); // true
            System.Console.WriteLine(y.HasValue); // false
            System.Console.WriteLine(y ?? 123);   // 123

            string z = null;
            System.Console.WriteLine(z ?? "foo"); // foo


ronnieholm:
Pull request created from master. Will resubmit one created from dev.

Issue #392, opened 2015-01-17
Do not check for literal exception message text


erwinvanhunen:
In the Core library there is code that checks for the actual exception message. E.g.:

if (ex.Message.StartsWith("Invalid field name. {33e33eca-7712-4f3d-ab83-6848789fc9b6}", StringComparison.OrdinalIgnoreCase))

This check will fail on those site collections where the default language is not English. We need guidance/examples on how to handle these cases correctly.

SharePointRadi:
Duplicate with #326. What is the best way to handle this?

jansenbe:
I'll close #326 with a reference to this bug.

jansenbe:
FYI: Have fixed this for tenant extensions. Others still todo

jansenbe:
Fixed most others...

PR #444, opened 2015-02-12
SiteExists method language agnostic implementation

nanddeepn:
Current implementation of SiteExists method works on sites created in EN language as the ex.Message check is performed against "Unable to access site" and "Cannot get site". Although non existing site throws error "File Not Found."
Removed the language specific site existence check. With new implementation:
1. Check if site exists
2. Check if site is recycled.
3. Return false, if 1 and 2 not satisfied.

### DIFF
--- a/OfficeDevPnP.Core/OfficeDevPnP.Core/AppModelExtensions/TenantExtensions.cs
+++ b/OfficeDevPnP.Core/OfficeDevPnP.Core/AppModelExtensions/TenantExtensions.cs
@@ -214,7 +214,7 @@ namespace Microsoft.SharePoint.Client
         {
             try
             {
-                //Get the site name
+                // Get the site name
                 var properties = tenant.GetSitePropertiesByUrl(siteFullUrl, false);
                 tenant.Context.Load(properties);
                 tenant.Context.ExecuteQueryRetry();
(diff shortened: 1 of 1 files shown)


jansenbe:
Hi Nanddeep, thanks for the proposed change.

We're not going to take this change as we're still looking into a better solution (see #392). The idea is to use server error versions or look at making the string translatable via resource files

nanddeepn:
Thank you for the update @jansenbe 

I had a look at the OOB Resources files in 15 hive on my farm to see if the errors are being picked from there, but I was wrong.
Please let me know, if I can contribute any part in your research because I being mostly working with non English sites development - I had to always look out for the language agnostic solutions. Hope you understand my pain :)

jansenbe:
I certainly do and we're working on it. Are you saying that you don't see these error messages in the 15 resource files?

nanddeepn:
Hello Bert,
Yes, I did not see the error message in 15 resource files. I will recheck if I have overlooked anything and confirm you by couple of hours.

nanddeepn:
Hello Bert,
I did not find the error messages in 15 resources files.

Found a few mentioned below, but no luck
Unable to access site
Cannot get site

PR #448, opened 2015-02-12
Minor edits to code comments.

Lauragra:


VesaJuvonen:
This code is coming from Visual Studio templates and should be fixed there. Otherwise you'd have to do this specifically for each and every solution which we create in PnP. This means that you could certainly do these changes, but this is endless road for you, which is not really recommended.

VesaJuvonen:
Reverted this due numerous reasons. Did not notice that this was for PnP... and we should not do this update for all solution, since this is oob tokenhelper.cs, which is coming from VS templates. PR was also targeted to master branch, which is not correct... all changes should be targeted to dev branch and then they are merged to master on monthly basis.

Lauragra:
Adding Reeza. This effort came about because as we prepared the articles for publishing to MSDN, we made updates to the code comments to conform to MSDN style. We then decided that we would update the code comments in the samples so that they follow the same style. Are you saying that there is a better way to go about this?

Thanks,
Laura

PR #458, opened 2015-02-17
Added search extension methods

nanddeepn:
Added extension methods for search.
1. ImportSearchSettings
2. ExportSearchSettings
3. SetSearchCenterUrl
4. GetSearchCenterUrl


nanddeepn:
ResultsPageAddress: Specifies the URL of the web page to which the search box redirects to display the results of a search.
Reference: https://msdn.microsoft.com/en-us/library/jj661921%28v=office.12%29.aspx?f=255&MSPPError=-2147217396

Since, we want to Set the search center url, we should pass null for this parameter (as we do not the search result to redirect to any page, rather to the search site for e.g. /Search).

jansenbe:
Thanks Nanddeep...i'm processing this pr

nanddeepn:
Thank you Bert!

PR #461, opened 2015-02-18
Samples showing how to include rollback logic in app event handlers

Rick-Kirkham:
One sample shows how to use the handler delegation strategy to minimize the number of CSOM calls from the handler web service back to SharePoint. It also shows how to use the CSOM ExceptionHandlingScope and ConditionalScope classes. The example task is adding a list to the host web in the AppInstalled handler and recycling that list in the AppUninstalling handler. The other sample shows how to do the same thing in scenarios where the handler delegation strategy cannot be used.


jansenbe:
Thanks Ricky, I'll be looking into these

jansenbe:
Nice samples, especially the one using the  CSOM ExceptionHandlingScope and ConditionalScope classes is not very well known and deserves its own sample.

PR #462, opened 2015-02-18
Rework how nodes are rendered in taxonomypicker

pbjorklund:
Tested with a termset of 7000 terms, loading the termset and rendering the termset take roughly the same time. But at least the waiting spinner works until it reaches the forEach loop now, and it only locks the UI thread for a few seconds instead of a few minutes now.

Performance in the UI is not pleasant, but thats because of the DOM structure generated. At least we got much more performant code now when I stopped relying on jQuery.append(). Inserting a few hundred thousand characters is going to take it's time.

Resorting to deferreds because forEach is not synchronous when you
run functions inside of a forEach call.

forEach is supported on IE9+ only. Problem?

I would really like some feedback on this because I think it's more complicated than it has to be, but this was the only way I could get it working.

### DIFF
--- a/Components/Core.TaxonomyPicker/Core.TaxonomyPickerWeb/Scripts/taxonomypickercontrol.js
+++ b/Components/Core.TaxonomyPicker/Core.TaxonomyPickerWeb/Scripts/taxonomypickercontrol.js
@@ -994,14 +994,11 @@
 
         outHtml += '<ul class="cam-taxpicker-treenode-ul" ' + addlStyle + '></ul>';
 
-        //var batchCounter = 0;
-
         var defs = [];
 
         termList.forEach(function (term) {
             var deferred = $.Deferred();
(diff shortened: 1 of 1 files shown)


pbjorklund:
I'l send over a partner agreement as soon as I can find my printer.

(Why no Office 365 Docusign? :smile: )

pbjorklund:
I made some gifs to illustrate the magnitude of the problem and change.

before
after

jansenbe:
Awesome work Patrick! Clever fixes for a though problem to solve :)

IE9+ is absolutely OK.

VesaJuvonen:
This is great fix indeed. Thanks for the great work.

VesaJuvonen:
"Why no Office 365 Docusign?" - We are working on better process, but we had to come up with something to start with based on legal request... it's only one time thing, so not that bad.

PR #467, opened 2015-02-21
Fix rendering of rootnode in taxonomypicker, remove uncessesary anon func

pbjorklund:
1. The root node did not render.

2. For some reason I called a anon func that could just be inlined in the
current function.

Bug: Adding nodes to anything but the root node does not work, i.e adding a child node does not work. Not really sure why this is, @jansenbe got any ideas?


jansenbe:
Don't really understand what goes wrong with adding child nodes. I've tested this and get them nicely added in IE11 and Chrome:

pbjorklund:
I'm talking about adding new childnodes to an open term set through the "add" link that is only displayed when the taxpicker is setup with open





On Sat, Feb 21, 2015 at 11:41 PM -0800, "Bert Jansen" <notifications@github.com> wrote:












Don't really understand what goes wrong with adding child nodes. I've tested this and get them nicely added in IE11 and Chrome:

PR #469, opened 2015-02-23
Core.SiteInformation

SharePointRadi:
This sample shows a few simple things combined together. It displays information about the current site collection. It makes use of custom actions, remote event receives and a few API calls both under site collection scope and tenant scope.


jansenbe:
Thanks Radi. Will look into this and merge

jansenbe:
Hi Radi, current version of this sample fails on the below line in the RER:

var appPageUrl = string.Format("https://{0}/Pages/Default.aspx?{{StandardTokens}}", System.Web.HttpContext.Current.Request.Url.Authority);

System.Web.HttpContext.Current = null in a RER call.

PR #484, opened 2015-02-26
Resolve #468 Components/Taxonomypicker - Empty UL was not inserted for childnodes.

### LINKED ISSUE #468: Adding notes to open termset fails when not adding to the root node
Adding new childnodes to an open term set through the "add" link that is only displayed when the taxpicker is setup with an open termset fails when not being added to the root node

pbjorklund:
Adding a new node requires that the containing div has a sibling UL that
is empty where the new node can be added. This UL was not constructed for
childnodes.

This resolves #468


jansenbe:
Thanks Patrick. I'll test it and merge

jansenbe:
Works perfect :-)

PR #485, opened 2015-02-26
Office365 API Demo - ASP.NET Web Forms

PaoloPia:
Hi guys,
I've just released the demo about how to use Office 365 API, together with OpenID Connect and multi-tenancy, within an ASP.NET Web Forms web application.


jansenbe:
Thanks for this Paolo.

When testing I bump into manager = null on this line:

var manager = Context.GetOwinContext().GetUserManager<ApplicationUserManager>();

Is there any special requirement?

PaoloPia:
Sorry for that ... it is an issue not related to the real focus of the demo, which are the Office365 APIs. BTW I worked around it in the PR that I'm going to submit. Thanks and good catch! :-)

jansenbe:
ok. I've merged this one for now, will process your PR with the fix once there. Another thing that I noticed is that having the statements  <add key="ida:ClientId" value="[put here your ClientID]" /> throws an error when you add a connected service is VS...maybe better to clean them completely?

PR #486, opened 2015-02-26
Added method to Publish a Custom Event in the WorkflowExtensions type.

PaoloPia:
I've added an extension method to support publishing custom events to workflow instances. However, as far as I can see, we miss a test plan for the WorkflowExtensions type. I would like to think about a suitable way to create a test plan for the WorkflowExtensions type, as like as we have for most of the other extension types. Nevertheless, it is not an easy task, because it involves having workflow definitions and workflow instances to play with. What's your feeling? Is It Worth the time, to find a way to test the WorkflowExtensions type? Thanks.

Of course, if you think that this method is useless ... just skip it :).


jansenbe:
Nice one, thanks Paolo

PR #499, opened 2015-03-04
Added overload SendEmail methods

nanddeepn:
Added overload SendEmail method that accepts password as SecureString, which is usual scenario.
Added SendEmail method for on-premise environment.


jansenbe:
Thanks Nanddeep

jansenbe:
Merged but did remove the compiler directives: both the SMTP sending as using the SharePoint CSOM SendEmail method do work for 15 and 16.

nanddeepn:
Good point Bert. Thank you.

PR #501, opened 2015-03-04
Only load resource files once for taxonomypicker

pbjorklund:
Assume that we only load one resource file per usage of taxonomypicker and store the state in the CAMControl object that is instantiated on running the anon function we use to start it up. Alert if loading of the resource file fails.

The resources does not necessarily have to live inside the taxonomypicker object, but that is a bigger change.


jansenbe:
Thanks Patrik. I'll have a look at this later today

jansenbe:
Tested and works fine...clean/clever solution...I like it :)

Issue #498, opened 2015-03-04
Setting default value to Texonomy Field 


shankargurav:
Hi,

We are in in need of setting default value to Texonomy Field. Currently PnP supports setting Texonomy Field default value at list level. We were looking to set default value at field level, but there is no provision to do this through CSOM. Wants to know is it even possible to set default value of Texonomy Field at field level?

Thanks,
Shankar Gurav

jansenbe:
#505 has been merged. Thanks @shankargurav

PR #505, opened 2015-03-05
Setting default value to Taxonomy field

shankargurav:
With reference to this https://github.com/OfficeDev/PnP/issues/498, I have made changes in TaxonomyExtensions.cs to have default value itself while creating taxonomy field.


jansenbe:
Thanks a lot for this contribution! 

Before we can accept your changes we would need to ask you to sign a [collaboration agreement](https://github.com/OfficeDev/PnP/blob/dev/Microsoft%20Corporation%20Collaboration%20Agreement-Officedev%20Patterns%20%26%20Practices%20GetHub%20Project.docx). Can you please fill in the sections highlighted in yellow, sign it and send a scanned version to Office365PnP_CA@microsoft.com. This agreement only needs to be signed once, so for all your future pull requests you're fine. Sorry for the extra hassle, but it is a new requirement from our legal department.

jansenbe:
Hi Shankar, I'll be looking into this.

jansenbe:
Thanks for your contribution. We've for now taken the method to set a default value and made that an extension method that you can call after you've created the taxonomy field. Extending the fieldcreationinformation class with a DefaultValue is something that we might look into in the future.

Issue #521, opened 2015-03-10
Tenant.GetSiteCollections only returns first 300


SharePointRadi:
As described here: https://www.yammer.com/itpronetwork/#/Threads/show?threadId=508865792
This needs to iterate by 300. 

I will tackle this quick fix.

SharePointRadi:
Fix is provided in #523, waiting for someone else from the core team to confirm changes.

jansenbe:
#523 has been merged. Thanks @SharePointRadi

PR #522, opened 2015-03-10
Fix GetSiteCollection #521

SharePointRadi:
Can someone please review for quality as this modifies the Core project :) 

Enhanced GetSiteCollections to honor the 300 item limit by the GetSiteProperties method.
Carefully added method overrides to not break signatures of existing solutions
The method now has 3 overrides where you can provide a startIndex and includeDetail to fetch other SPSite properties.
The API does not provide a Take() equivilent though...


SharePointRadi:
I hope I didn't screw something up here... I accidently made this pull request to the master branch instead of the dev branch, then panicked and closed it. Was that the right thing to do? I couldn't find a way to edit it.

jansenbe:
Also ensure you create the PR against the DEV branch.

PR #541, opened 2015-03-17
Added method to set page output cache

nanddeepn:
Added method to set the page output cache to improve the page load time, based on pre-existing cache profiles.


jansenbe:
Thx Nanddeep. I'll be looking into this.

PR #577, opened 2015-03-26
Added support for selecting sugestions using the keyboard, and some smal...

rjbooden:
...ler updates. Smaller updates: Added support for enter key. Getting the marker more reliably, as doing a backspace, after a new keyword was added, resulted only in removing the ; you only were able to remove further text after clicking in the taxonomy field. Only allowing for one new taxonomy value to be added at once. Keywords only allow one level, when adding a second level keyword, it disappears when entered, now you can't add a second level keyword, and will instead add one to the root. Adding a new taxonomy value and then adding one in the next level on the new taxonomy node made possible by adding missing ul. The dlgCurrTermNode should be the newly created title node.


jansenbe:
Thanks for this contribution Ronald. I'll be looking into this.

jansenbe:
Been testing and works perfect...gives us nice keyboard support for the taxonomy control. Really appreciated :-)

PR #621, opened 2015-04-03
Use a cached for loop instead of forEach to resolve #614

### LINKED ISSUE #614: Core.TaxonomyPickerWeb - taxonomypickercontrol.js - forEach function not compatible in IE8
Hi,

Its an excellent component. Many thanks for sharing.

While I was using this component in IE8, I came across an error, which is the modal pop up opens up with no terms in tree view.

On my investigation. 

Issue:
Function "buildTermSetTreeLevel" in taxonomypickercontrol.js uses forEach to loop around termList and create a tree structure. IE8 does not support forEach function, hence it causes error.

Resolution: 
There can be many solutions to it but I am mentioning 2. 
1) We can use jQuery.each instead.
Or
2) We can add a prototype for forEach in project initiating scripts allowing use of forEach() in implementations that don't natively support it
prototype code can be referred at below link
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach

pbjorklund:
Resolve #614 by using a for loop that caches length.

There is some blocking of UI at the end of the load, but that is when inserting the extremely large new DOM element.

### DIFF
--- a/Components/Core.TaxonomyPicker/Core.TaxonomyPickerWeb/Scripts/taxonomypickercontrol.js
+++ b/Components/Core.TaxonomyPicker/Core.TaxonomyPickerWeb/Scripts/taxonomypickercontrol.js
@@ -1009,7 +1009,8 @@
 
         var defs = [];
 
-        termList.forEach(function (term) {
+        for (var i = 0, len = termList.length; i < len; i++) {
+            var term = termList[i];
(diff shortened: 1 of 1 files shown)


erwinvanhunen:
Thanks Patrik!

PR #656, opened 2015-04-14
Added support for a pages element in the provisioning model

erwinvanhunen:
See ProvisioningTemplate-2015-04-Sample-02.xml for an example on how the use the new pages element.


levalencia:
Amazing, just what I was looking for, cant wait to see it in the Main branch and the nuget package.
I have a question though:

1. In the provided xml example you use rows and columns, I suppose thats for wiki pages only, what about webpart pages where you normally use zone names to know where the webpart will be placed.


Thanks a lot for this, great effort from the team

PR #662, opened 2015-04-14
Yammergroupfix

kendomen:
When "Group" is not selected, the toggle for the other two options should be disabled.


jansenbe:
Thanks Ken!

PR #663, opened 2015-04-14
Added support for adding webparts to webpart pages

erwinvanhunen:
Updated webpart element in schema to allow for Zone and Index parameter to facilitate provisioning web parts to existing webpart pages.
Made layout an optional attribute on the page element allowing for specifying webpart pages (to deploy a webpart page, use the File element in the schema)
Fixes some small issue with creating new wikipages through the Page element.


levalencia:
awesome!

PR #715, opened 2015-04-24
Added EnableMinorVersions property for ListInstance

vlad0:
Added EnableModeration(content approval) property for ListInstance
Updated .Equals() and .GetHashCode() of ListInstance
Updated XSD

EnableAttachments fix as it's not supported for DocumentLibraries and Surveys


jansenbe:
Thanks Vlad!

PR #717, opened 2015-04-24
Sample App for Outlook: UPS package tracker

dougperkes:
Another code sample, this time for Outlook. Demonstrates how to extract information from body of email, access a JSON API and format results in the app.


jansenbe:
Great Doug. I'll have a look at

PR #729, opened 2015-04-27
Power shell import term set

sgryphon:
Add PowerShell commands to import a term set using the standard format CSV file (the one you can download from Term Store admin as an example). Also added commands to get (to check if it exists) and create new Term Groups.

Also fixed a bug in the PSH doc generator that was duplicating syntax elements.


erwinvanhunen:
Thank you very much for this work! 

I've merged it manually due to some additional changes I wanted to do. I've renamed some of the cmdlets by removing Taxonomy, e.g. Import-SPOTaxonomyTermSet became Import-SPOTermSet.

I added a rooted path check in that cmdlet too.

PR #734, opened 2015-04-27
Dev

braegelno5:
added lines to disable 'Mobile Browser View' feature on web in InjectResponsiveCSS/Program, so the sample is now complete.


braegelno5:
the lines are compiling and the code throws no exception

jansenbe:
Thanks for this Torsten!

PR #737, opened 2015-04-28
a fix that allowes to add fields to existing content types

maratbakirov:
Imagine you have and xml file with content types and fiieldrefs and an existing system and you want to add the filedrefs added in the xml.

this simple fix allowes to do that without recreating content types 

note that overwrite does not work as it deletes the content type and creates it again but if the content types is used anywhere it cannot be deleted.

Hope this request can be accepted


jansenbe:
Hi Marat, thanks for this change. Being able to deal with "delta" processing is something we certainly want to add to the engine and when we do your code will be used. We however are not immediately going to merge this change as we want to offer delta processing for all the objecthandlers, not just for content types. Plan is to further work on this post the May release targeting the June release (or later).

maratbakirov:
Could we make this fix before the full delta engine is introduced? The good thing is it does not change anything in XML format or external interfaces or calls. It just mades the whole thing a little more usable. 


I am so sorry to say that but for now the XML engine without this fix is not so usable for us. An idea to delete a content type each time to just update it is not an option as then we have to recreate a lot of other dependand stuf..

Thank you!

jansenbe:
Hi Marat. This will not be in right now, sorry for that. What you could do is use the PnP provisioning engine extensibility model and write your own code that the engine will execute after it's done. See the tests to learn more on how this works: https://github.com/OfficeDev/PnP/tree/dev/OfficeDevPnP.Core/OfficeDevPnP.Core.Tests/Framework/ExtensibilityCallOut

PR #738, opened 2015-04-28
Bring TaxonomyPicker sample and Component in sync

pbjorklund:
Update sample and component to be in sync. In order to do
this I reimplemented the Samples functionality to accept
context, callbacks, LevelToShowTerms and Filter.

There is an option in the sample to use term as root node which
I did not reimplement, I don't really see the point of it and
it's not really trivial to do.

I did not update the README's of the projects, it warrants
a discussion of why there are two (now in sync) things in PnP
that does the same thing. Why not remove the sample?

### DIFF
--- a/Components/Core.TaxonomyPicker/Core.TaxonomyPickerWeb/Scripts/app.js
+++ b/Components/Core.TaxonomyPicker/Core.TaxonomyPickerWeb/Scripts/app.js
@@ -32,9 +32,8 @@ $(document).ready(function () {
                         function () {
                             $.getScript(layoutsRoot + 'sp.taxonomy.js',
                                 function () {
-                                    //binf the taxonomy picker to the default keywords termset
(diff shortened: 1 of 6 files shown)


jansenbe:
Thx Patrick. Will process this one today

jansenbe:
Have been investigating and seems there's a slight difference between this one and the component one. This version supports filtering of terms and supports to only show certain level in a hierarchical termset. These capabilities are not in the component version. 

Ideal scenario is:
*/ take over these capabilities to component version
*/ Update the documentation of the component version to highlight this capability
*/ Drop the sample 

To get a clean view on what was added to implement the filtering and hierarchical support you can check history of both files, get their initial commit version and compare them.

Do you want to take this Patrick?

pbjorklund:
They are in the component version as well now. This PR update both sample and component

PR #739, opened 2015-04-28
List instance fixes and extensions

vlad0:
OnQuickLaunch fix - the ListCreationInformation object doesn't set the List OnQuickLaunch property.
RemoveExistingViews - throws exception "The last content type on a list cannot be deleted." Added try catch and description of the error. Need to be handled properly.
OnQuickLaunch & RemoveExistingViews Collision fix - when RemoveExistingView="true" and OnQuickLaunch="true" the List OnQuickLaunch property is set to false. When new views are added to the list the OnQuickLaunch property should be re-set.


erwinvanhunen:
I noticed that you merged the dev branch after you committed the "default contenttype binding should be set to last because...". That effectively caused your commit to disappear. Was that on purpose?

vlad0:
Yes it's on purpose. By the time I've pushed someone has already fixed and pushed it to the dev branch. My commit became useless and the merge was a must :)

erwinvanhunen:
That was actually me. However, your finding about the default CT still stands I think. The additional attribute I added simply sets the default Ct straight after it finds a CT in the list of CTs to add with a Default property set to true. E.g. the unique list of CTs is still not reflecting the possible complete list of CTs. I reintroduced your fix and merged it my fix. Will merge soon.

PR #743, opened 2015-04-28
Update location of credential prompt in Program.cs

skaggej:
Relocated the prompt for credentials from lines 49-51 to lines 30-32 and added line 15 as a result of the move.  This simplifies the execution of the program by only prompting the executor for credentials once.  Prior to this change it was found that the executor would be prompted for credentials for each site collection configured within the settings.xml file.


jansenbe:
Thx Eric.

PR #764, opened 2015-04-30
site provision onprem fix

ivanvagunin:
Fix site provisioning engine crash. ClientQueryable in CSOM v15 does not implement GetEnumerator, so linq queries against collections of client objects throw exceptions


jansenbe:
Thanks Ivan. I'll have a look at this.

Issue #763, opened 2015-04-30
Token Helper is inconsistent


maratbakirov:
PnP-dev\OfficeDevPnP.Core\OfficeDevPnP.Core\Framework\Provisioning\ObjectHandlers\TokenDefinitions\SiteCollectionToken.cs

and

PnP-dev\OfficeDevPnP.Core\OfficeDevPnP.Core\Framework\Provisioning\ObjectHandlers\TokenDefinitions\SiteToken.cs

for the root site (i.e. http://xx.sharepoint.com/) they return "/"

for not root site they return the path without ending slash

http://xx.sharepoint.com/sites/aa) 
the value will be "/sites/aa"

this leads to inconsistency when using the tokens

for example we do not know if we shoud write ~sitecollection_catalogs/masterpage/layout.aspx

or 
for example we do not know if we shoud write ~sitecollection/_catalogs/masterpage/layout.aspx

I suggest to make it consistent and always not to have the trainling slash (it is easier to add the slash than to remove it so it is better to have a token without the trailing slash)

maratbakirov:
Suggested change 
https://github.com/OfficeDev/PnP/commit/fa957bd341177214286adbef02c68b6b5cd41005

I am not issuing a pull request as it will suggest my other changes to issues 762 and 747

https://github.com/OfficeDev/PnP/issues/762
https://github.com/OfficeDev/PnP/issues/747

and I am not sure if 762 is going to be approved.

erwinvanhunen:
That's a very valid argument you're making! I'll fix it.

erwinvanhunen:
Merged with PR #766

erwinvanhunen:
I will close this issue now

maratbakirov:
I have posted a pull request as well:) might be conflitcts. thanks for quick fix though!

Issue #762, opened 2015-04-30
Could it be possible to support tokens like ~sitecollection in file properties for new provisioning engine?


maratbakirov:
new xml provisioning engine

I am provisinling a file and need to specify a PublishingPageLayout property 

with the Value

~sitecollection/_catalogs/masterpage/BlankWebPartPage.aspx, Blank Web Part Page

could be good if we could support ~sitecollection token replacement in file properties, it is not working now.

maratbakirov:
suggested change (very simple)
https://github.com/OfficeDev/PnP/commit/c441d7c829e6c18aaf0bb3812e0d72e81225c802

I cannot shoot a pull request as it will also suggest my other change that is not approved yet.

PR #765, opened 2015-04-30
Dev

maratbakirov:
https://github.com/OfficeDev/PnP/issues/762
https://github.com/OfficeDev/PnP/issues/763


erwinvanhunen:
Thank you this PR! Good additions indeed. I've implemented them, but a bit differently. You proposed to update the file property parsing and you modified the FileFolderExtensions.cs file for that. However, that's a core library class, which not necessarily knows about the tokenparser static class. E.g. it might not be initialized. I've implemented your proposed functionality in the Files Object Handler (ObjectFiles.cs) of the engine now, as that one is aware of the TokenParser.

I've also updated the site and sitecollection token to strip the trailing slash.

See PR #767 and PR #766 for the implementation.

I will close this PR now.

PR #781, opened 2015-05-03
Fixes for viewport meta tag

StfBauer:
Viewport meta tage code added to provisioning updated readme

### DIFF
--- a/Samples/Branding.InjectResponsiveCSS/Branding.InjectResponsiveCSS/Program.cs
+++ b/Samples/Branding.InjectResponsiveCSS/Branding.InjectResponsiveCSS/Program.cs
@@ -31,7 +31,10 @@ namespace Branding.InjectResponsiveCSS
                 ctx.Credentials = new SharePointOnlineCredentials(userName, pwd);
 
                 Web web = ctx.Web;
+                var allProperties = web.AllProperties;
+
                 ctx.Load(web);
+                ctx.Load(allProperties);
                 ctx.ExecuteQuery();
 
                 UploadAssetsToHostWeb(web);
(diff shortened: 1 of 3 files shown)


jansenbe:
Thx Stefan. If you can still sign the CLA before tomorrow than we can take this PR still in the May release which will be created tomorrow.

StfBauer:
Can you lend me a hand how to do this? :)

jansenbe:
Follow the steps from here: https://cla.microsoft.com.

StfBauer:
Thanks and done.

jansenbe:
FYI: still need to test this, but does this change also work for non-publishing sites?

StfBauer:
Yes the only thing that needs to be done is to enable publishing infrastructure but no other related publish site feature. The publishing infrastructure enables the SEO Features that will be used to add a proper view port. I tested it with team site site collection and publishing site collection. If those properties doesn’t exist it simply skips the code. Another option is to add the viewport with javascript inside a custom action. This is currently not covered. I updated the readme file with those informations too.

StfBauer:
Update: You can set the properties but those will be ignored on non-publishing page. So for non-publishing pages you need to use a custom action. Without a proper viewport the RWD won’t work anyway on non desktop browser.

VesaJuvonen:
In best case scenario we should get this included without the demand of publishing feature activation, since we do not really recommend that to be done for oob team sites due possible impact in long run. You could however also add the needed metadata entry using JS Embed model and with some jQuery. That means that no changes to oob master pages is required and there would not be any dependency on publishing feature.

StfBauer:
From the pattern perspective I agree with you we should remove the thing from the read me but keep a side note for publishing features (Publishing Portal / Enterprise Wiki) enable site the SEO option can be used. While all other site templates should embed it with JavaScript. To be honest. The use case was built based on an enterprise wiki. So there are much more things that cause problems on team sites with this css. On problem for example is that the text layouts in the team site are just semi-responsive.

PR #787, opened 2015-05-05
fixes - no default termstore and extensibility assembly name

ivanvagunin:
Small fixes for site provisioning engine:
1- OnPrem token parser crashes when no default term store or no MMS
2- Template parsing ExtensibilityProviders, provider assembly name set to AssemblyQualifiedName (which is type name) instead of assembly name, so ExtensibilityManager cannot create instance of provider


erwinvanhunen:
Thanks for these fixes, Ivan! I'll merge them.

PR #788, opened 2015-05-05
Adding Support to Core for Enabling Ratings/Likes 

akhileshnirapure:
Hi Team,

I've ported my sample code from **Core.ListRatingSettings** and exposed ***SetRatings()*** extension method on List Object (in a partial class). 

Also added supporting test scenario in Test Project (ListRatingExtensionTest)

**TODO**: I yet need to add appropriate logging.

Thanks
Akhilesh


biste5:
Hi Akhilesh, 

Massimo here.

I've had a look at your PR, and I have some feedback around that, and interesting related scenario to implement
Feedback about your PR is that you should look better at fast-forwarding your local branches, in order to include only commits related to your changes, and avoid local merges every time, which generate 'commit branch...' commits. These make merging work for Bert and others more difficult, and also harder to follow revisions history.
Check out [this](https://github.com/OfficeDev/PnP/wiki/contributing-to-Office-365-developer-patterns-and-practices) page again, and if you need help, you know how to find me :)

There are other functionalities in SharePoint which can be enabled in this way. 
For what I can recap:
- Enterprise Keywords (TaxKeyword)
- Geolocation: this is a bit different, because creates a new column getting input information as well

I would love to see this implemented in core, and PowerShell commands too!

Massimo

akhileshnirapure:
Hi Massimo,

Thanks for feedback, but i 2nd your thought, Git is quite intelligent in merging most of the time you don't really need to compare a lot (if commit is done with the synced copy) and as far as changes are done in new files they are even quite easy. My change is pretty straight forward and they are in new files.
I agree if the case is modifying existing code it can become bit tedious.

Thanks for suggestions, i already thought of TaxonomyKeyword, but not for GeoLocation field it would be a good add too.

Thanks
Akhilesh

jansenbe:
Thanks Akhilesh. I'll have a look into this

PR #789, opened 2015-05-06
Created a simple console app to make testing site provisioning easier…

SharePointRadi:
… and to illustrate "getting started" with site provisioning.

It really is just to quickly demonstrate it to new people. I will also use it for a "getting started" blog post.

PR #790, opened 2015-05-06
Core fixes

SharePointRadi:
I've added a few checks that are causing my provisioning to fail on a demo Contoso O365 site.

The try/catch statements log errors, which I visualize in red in the Core.SiteProvisioning.Console sample.


SharePointRadi:
I'm not sure I did this right, my intent was to create 1 pull request for fixes to the core, and another for a sample. They are in different branches.

SharePointRadi:
Closing in favor of #789

PR #791, opened 2015-05-06
XML Documentation

joshuapetryk:
Building with XML documentation would greatly enhance the extensions with intelliesense.


jansenbe:
Good point Joshua! Thanks for fixing this.

PR #793, opened 2015-05-08
Provisioning engine fixes

mwiedemeyer:
This PR includes the following fixes/changes:
- `EnableFolderCreation` was always true even for lists which don't support folders
- Fixed an issue when a site has no composed looks
- Updated provisioning scenario console app to support OnPrem environments

I hope it is ok to have these three changes in one PR.


mwiedemeyer:
Sorry, the PR was made against the master branch and not the dev branch

PR #794, opened 2015-05-08
Provisioning engine fixes

mwiedemeyer:
This PR includes the following fixes/changes:

- `EnableFolderCreation` was always true even for lists which don't support folders
- Fixed an issue when a site has no composed looks
- Updated provisioning scenario console app to support OnPrem environments

I hope it is ok to have these three changes in one PR.

### DIFF
--- a/OfficeDevPnP.Core/OfficeDevPnP.Core/AppModelExtensions/BrandingExtensions.cs
+++ b/OfficeDevPnP.Core/OfficeDevPnP.Core/AppModelExtensions/BrandingExtensions.cs
@@ -920,6 +920,11 @@ namespace Microsoft.SharePoint.Client
                 }
             }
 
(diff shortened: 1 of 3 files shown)


erwinvanhunen:
Hi Marco!

Thank you for your contribution! I will check it out later today.

/Erwin

erwinvanhunen:
Custom domains in O365 can only be used for the public site, which is also addressible as <tenantname>-public.sharepoint.com. The public site functionality will be removed in the near future, I don't think we should worry too much about those cases, as it can be solved by addressing the site with the other url ending on sharepoint.com.

mwiedemeyer:
Ok, I've updated the check.

erwinvanhunen:
Thanks! Merged.

PR #800, opened 2015-05-11
Provisioning schema extensions

vlad0:
- Added ListInstance DraftVersionVisibility property to the provisioning schema
- Added extension for setting WelcomePage of a web by ServerRelativeUrl
- Added option for provisioning schema to set a page as a WelcomePage of the web.RootFolder


OfficeDevPnP.Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectPages.cs
```diff
@@ -69,6 +69,12 @@ public override void ProvisionObjects(Web web, ProvisioningTemplate template)
                     web.AddWikiPageByUrl(url);
                     web.AddLayoutToWikiPage(page.Layout, url);
                 }
+
+                if (page.WelcomePage)
+                {
+                   web.SetHomePageByServerRelativePath(url);
```
erwinvanhunen: why not use the existing SetHomePage extension method?

vlad0: The "url" variable is a server relative url. 
 if (!url.ToLower().StartsWith(web.ServerRelativeUrl.ToLower()))
                {
                    url = UrlUtility.Combine(web.ServerRelativeUrl, url);
                }

.SetHomePage() doesn't work with Server Relative Url but Web Relative Url. That's why i added new extension.

erwinvanhunen: I understand your reasoning. I think though that we shouldn't add new extension methods because one specific scenario requires it. It shouldn't be too complex to create a rootfolder relative url from the server relative url in the object handler and then use that one instead?

vlad0: Probably you are right :) The logic for setting the WelcomePage in the ObjectHandler or in the BrandingExtensions is the same. My thoughts has been that being in the BrandingExtension makes it reusable, otherwise it lives only in the ObjectHandler waiting for copy/paste driven development. Last but not least I cannot be sure someone gonna need it as extension. Please advise how to proceed :)

erwinvanhunen: I would just convert the serverrelative url to a folder relative url and then send that to the existing sethomepage method and not introduce a new extension.

< untested code >
var rootFolderRelativeUrl = url.Substring(web.RootFolder.ServerRelativeUrl.Length);
web.SetHomePage(rootFolderRelativeUrl);
</ untested code >

Sorry for nitpicking. I do think the rest of the PR is absolutely valid and good additions :-)

vlad0: Hi Erwin,

I have just committed the way you proposed. Thanks again for the review! :)

jansenbe:
Nice work Vlad. Can you complete the CLA registration (https://cla.microsoft.com)...sorry for the extra hassle (Radi knows how to do this)

erwinvanhunen:
Thanks Vlad, I'll work on this one.

PR #804, opened 2015-05-14
Allow DataRows in separate template for 2nd pass

dmdj:
We really needed ability to apply DataRows *after* Lists are created and Lookup columns resolved (since we are populating Lookup columns in some cases).  The template now drives the application of DataRows - regardless if it is part of the initial List creation template or a separate template.

### DIFF
--- a/OfficeDevPnP.Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectListInstance.cs
+++ b/OfficeDevPnP.Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectListInstance.cs
@@ -322,14 +322,20 @@ namespace OfficeDevPnP.Core.Framework.Provisioning.ObjectHandlers
                 #endregion
 
                 #region DataRows
-
-                foreach (var listInfo in createdLists)
+                
+                //Allow DataRows from separate Template (after List instances created and Lookups have been resolved)
+                foreach (var listInstance in template.Lists)
                 {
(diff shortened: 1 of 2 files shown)


erwinvanhunen:
Hi David! 

Thank you for your contribution! I will check it out later today.

/Erwin

dmdj:
I'm going to change this slightly and re-submit such that it only adds DataRows when the list is initially empty.

dmdj:
I added check for ItemCount so data is only provisioned to empty lists.

erwinvanhunen:
I had a look at your code. I actually think we should handle this differently. In the solution you'll find a ObjectLookupFields.cs class which is run after the lists are created. The ObjectLookupFields class will fix/enable all lookup fields that have been deployed through the template. I would say that we move the code to create the list items either to the end of that class, or we introduce a new object handler class that we run after the ObjectLookupFields handler.

When it comes to the ability of adding list items to a list that is being created in a separate template: that will be handled automatically the moment we implement delta handling (= functionality to update existing artifacts). We want to make sure that we implement delta handling for all artifacts items in one go, to provide a consistent behavior of the engine, which is one of the reasons it's not there yet in the engine.

dmdj:
Yes, that makes good sense.  Let me take another shot at it with the focus of applying DataRows from a single template.

erwinvanhunen:
Proposal: rename ObjectLookupFields.cs to ObjectPostListProcessing.cs (or something alike) and move the list item creation code to that one.

PR #810, opened 2015-05-15
Add JSLINK Announcements list Accordion

rovcaene:
I added a sample that transforms an announcements list into an accordion. This sample illustrates the use of the OnPreRender, OnPostRender, Header, Item and Footer jslink possibilities.


jansenbe:
Nice work from Belgium Robrecht :)

Will have a look and merge, thx!

PR #832, opened 2015-05-22
Bug in ListExtensions with logging

arnoldboersma:
Updated the logging of the createlist, Logging source was missing as parameter as a result the logging doesnt work as expected.


jansenbe:
Thanks Arnold. Will process this PR later today

PR #840, opened 2015-05-23
Fixes logging resources references for adding a site.

NickDarvey:


jansenbe:
Hi Nick,

Thanks for this change, this was indeed a little error...but it was already fixed. You might not have noticed since you created a PR against the master branch. The way we work is that we accept PR's against the dev branch, but since in this case this change is already in dev I'm gonna simply close the pr.

PR #843, opened 2015-05-25
Fixed Stream copy bug & spell correction

chirag64:
CopyTo fails to copy stream data because Position is not set to 0


erwinvanhunen:
Thanks!

PR #849, opened 2015-05-27
Make GetFile handle exception type instead of english exception message

pbjorklund:
Change from
```if (ex.Message == "File Not Found.")```
to
```if (sex.ServerErrorTypeName == "System.IO.FileNotFoundException")```


jansenbe:
Thanks Patrik

PR #850, opened 2015-05-28
Update to check illegal characters

ZeBobo5:
Use the available method extension


jansenbe:
Thanks. Will process this PR

PR #860, opened 2015-05-31
fix to create Lists in xml template and some more refactoring

braegelno5:
on ShareCamp at Munich we have found that there was an error in CreateEntities to check if the list we have found in the web should written to the xml template. if (index == -1) {}
changed to if (index != -1) {} and then refactored

### DIFF
--- a/OfficeDevPnP.Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectListInstance.cs
+++ b/OfficeDevPnP.Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectListInstance.cs
@@ -1,13 +1,9 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Data.SqlTypes;
 using System.Linq;
 using System.Text;
-using System.Web.Configuration;
-using System.Web.Instrumentation;
 using System.Xml.Linq;
 using Microsoft.SharePoint.Client;
-using OfficeDevPnP.Core.Framework.ObjectHandlers;
 using OfficeDevPnP.Core.Framework.ObjectHandlers.TokenDefinitions;
 using OfficeDevPnP.Core.Framework.Provisioning.Model;
 using OfficeDevPnP.Core.Utilities;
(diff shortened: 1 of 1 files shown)


braegelno5:
it now writes pnp:Lists elements in our xml template again

erwinvanhunen:
I'm a bit confused by this PR. This is currently in the dev branch:

```C#
int index = -1;
if (creationInfo.BaseTemplate != null)
{
    // Check if we need to skip this list...if so let's do it before we gather all the other information for this list...improves perf
    index = creationInfo.BaseTemplate.Lists.FindIndex(f => f.Url.Equals(item.RootFolder.ServerRelativeUrl.Substring(serverRelativeUrl.Length)) &&
                                                                  f.TemplateType.Equals(item.BaseTemplate));
}
if (index == -1)
{
```

This is in your PR:

```C#
var index = -1;
if (creationInfo.BaseTemplate != null)
{
    // Check if we need to skip this list...
    // if so let's do it before we gather all the other information for this list...improves perf
    index = creationInfo.BaseTemplate.Lists.FindIndex(f => 
                        f.Url.Equals(item.RootFolder.ServerRelativeUrl.Substring(serverRelativeUrl.Length)) &&
                        f.TemplateType.Equals(item.BaseTemplate));
}
if (index == -1) continue;
```

That code would mean that the moment the list is not in the base template it will not be written to the template... Which is exactly the opposite of what we want to accomplish?

braegelno5:
At the moment, if I have at least one doc lib in my source site, the index is reset to a value greather than -1. But the if statement checks against the std -1, so we get no list in template XML.

erwinvanhunen:
what is the URL of the document library?

braegelno5:
../Shared Documents/

erwinvanhunen:
Then engine is behaving as designed. That is a built-in document library that comes with the STS#0 template and subsequently it will be filtered out of the generated template. We're working on implementing 'delta handling' which means that you can update existing artifacts (adding fields to existing lists, etc.), but right now, if an artifact like a document library is present in a out of the box site collection created from the STS#0 template we will filter it out. To do that we generated so called 'BaseTemplate'. E.g. we created sitecollection based on a site definition like STS#0. Then we generated a provisioningtemplate from that one, and included that template in the engine. When creating a template we find the corresponding base template and compare against that base template. If we have a match (index != -1) then we _don't_ export that artifact.

You can find the base templates we have in the OfficeDevPnp.Core\Framework\Provisioning\BaseTemplates folder.

jansenbe:
Hi Torsten,
Thanks for your fix. Depending on how one looks at the engine your solution could be the correct one but we've opted, as explained by Erwin, to filter out all the "default" elements that are part of the base template. We do this to get clean templates and we assume that in most cases people apply a template on top of the same site type as from which they've generated the template. We're still lacking documentation that explains this clearly, but we're working on that.

So I'm going to close this PR. Feel free to reach out to me or anyone else from the team if you would like to further discuss this.

braegelno5:
Ok i accept your decision, My thinking was for this community event at http://sharecamp.de/ to build a  custom site column add this to a ct and add this ct to an existing list. In this case the documents lib. But the lib wasn't  in theprovisioningtemplate  anymore..
After my changes the template was created with all the artifacts to create the lib in the target correctly. The only pitfall was here the check for list existence in the ProvisionObjects. So i deleted the lib in the target and the engine was able to show his working for the audience..

At the moment the engine checks twice:
1) exists list in source base template (STS#0) -- don't save anything to the provisioningtemplate
2) exists list in target on site rel url -- do nothing (or not implemented yet --do a merge)
To understand this by myself completely.. there are missing information's..
If this behavior is carved in stone we must check on all steps for changes:
1) check also for changes on an existing list from base template and save this to the template
2) if list exists check for diffs and merge them
Is this the plan to go?
For my understanding it would be easier to save all on the source side, than in the target do a create if not exist, do a merge on existing artifacts and delete elements from target if they not exists in the source anymore..

braegelno5:
Hi Bert,
i was coming to the conclusion, that the engine only persists changes outside from the initial template. But is a user working in that way, sometimes yes but not in general. 
What i'm missing on all Objecthandlers::CreateEntities on that if (check exists on initial template -- delete them from internal collection) is an else [ (check for changes on artefact -- save change to template) ].
If it's not there and at the moment, because there is no implementation, it's empty, my question would be, where will it in the future planned to host?
I would suggest to write a comment on the ObjectHandlerBase XML summary how the planning is. So it's all documented in the code;)

jansenbe:
The idea is indeed to proceed delta handling...this would mean delta handling for both the site creation as for getting a site. So if you create a team site and add a content type to the default "shared documents" library then this change should be picked up and applied. But as you know this is still work in progress and will take a while as all of this is driven by community work.

PR #864, opened 2015-06-01
to prove and check the proper way for delta list processing in ObjectHandler for ListInstance

braegelno5:
This changes nothing on the actual behavior of the engine!
The ObjectListInstance::CreateEntities is splited into two functions: AddNewListToTemplate and AddDeltaListToTemplate. The first hold all the actual working code if list not exist in 'BaseTemplate'. The second one writes only a warning to the log. 
Same as ObjectListInstance::ProvisionObjects, but here exist now a 'else' to handle future list delta processing, if list from provisioning template exists in target web. Writes only a Log.Warning.
I have done some XML comments as decorations to the functions, and i hope my way to step would not be again complete wrong;)

### DIFF
--- a/OfficeDevPnP.Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectListInstance.cs
+++ b/OfficeDevPnP.Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectListInstance.cs
@@ -1,13 +1,9 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Data.SqlTypes;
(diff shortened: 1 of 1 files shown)


erwinvanhunen:
Thank you for your contribution!

This is pretty much in line with how we are going to handle delta handling. I don't think we will merge your proposal right now as we're working towards a master merge in a few days, but after the release we will full-force work on delta-handling to be released for as much as possible object handlers in the next master merge. I already have some code ready to be merged there too.

braegelno5:
ok i will close this now, and will deeper thinking on the delta-handling

PR #872, opened 2015-06-02
Removed Version='1' from Field Definitions

ScotHillier:
Deployment fails when Version is included in field definitions.


jansenbe:
Thanks Scott. I'll process this PR

PR #878, opened 2015-06-03
save a changed list also to provisioning template

braegelno5:
i have only implemented now a  create entity for changed lists, where list def differs from base template list
on provision side i have added a List to hold ListInfo for updating, in the else path for existing lists is nothing implemented yet


jansenbe:
Hi Torsten, as discussed over IM we're closing this PR for now. We'll take in account your ideas around delta processing when we add delta handling for all object handlers...so already big thanks for providing these ideas.

PR #887, opened 2015-06-04
BaseTemplateTest changed to get test url's from App.config

braegelno5:
BaseTemplateTest hard coded url changed to App.config settings


jansenbe:
Hi Torsten,
Thanks for making me aware of these hard coded references...they've slipped in. Appreciate your attempt to parameterize the tests, but since these tests are anyway not correct I've decided to simply drop them. This means I can't merge your pr, but I'll ensure you get credited in the monthly release notes for the June release.

Issue #896, opened 2015-06-08
Inject CSS into ScriptBlock Custom Action - O365


werocool:
I am testing a senario where I provision a CustomAction for adding a custom css (rather than provision a CompositeLook or a Custom Masterpage) to a site collection. 

If I add token ({sitecollection}) to the scriptblock section, currently the code do not replace the TokenDefinition (ObjectCustomActions, line 77)
![inject](https://cloud.githubusercontent.com/assets/4926713/8040072/b430550a-0ddb-11e5-8b0d-bdc05d4e0728.png)

aaclage:
Did you try to use ScriptSrc="~SiteCollection/Style Library/PnP.css" and not ScriptBlock, scriptblock has a different time of loading in the Page life cycle.

werocool:
Yes, you are correct for the script part but the main branch was not replacing token (had a null exception when using in both scriptblock or scriptsrc). In the end, I ended up fixing the code (pull request accepted by Erwin) and then changed the css part to be:

https://github.com/OfficeDev/PnP/pull/898 

<pnp:Provisioning xmlns:pnp="http://schemas.dev.office.com/PnP/2015/05/ProvisioningSchema">
  <pnp:Preferences Generator="OfficeDevPnP.Core, Version=1.2.515.0, Culture=neutral, PublicKeyToken=null" />
  <pnp:Templates ID="BRANDING-RAC">
    <pnp:ProvisioningTemplate ID="BRANDING-RACDEPLOY" Version="1">
      <pnp:Files>
        <pnp:File Src="RAC.css" Folder="/Style Library/" Overwrite="true" />
      </pnp:Files>
      <pnp:ComposedLook Name="Office" ColorFile="" FontFile="" BackgroundFile="" MasterPage="{masterpagecatalog}/seattle.master" SiteLogo="{site}/SiteAssets/RAC-icon.png" AlternateCSS="{site}/Style Library/RAC.css" Version="0" />      
    </pnp:ProvisioningTemplate>
  </pnp:Templates>
</pnp:Provisioning>

aaclage:
ok i saw it now, thanks

PR #898, opened 2015-06-08
Issue #896

werocool:
In order to use tokens within the scriptblock of the customaction.


erwinvanhunen:
Thank you for your contribution! I will merge your code and add some small updates. It's important to rebase the tokenparser back to it's original context after it has been rebased earlier (TokenParser is a static object).

erwinvanhunen:
Some small remark: you're using a C# 6.0 feature on line 56. I reverted that back to the original code as not everyone has access to C# 6.0 yet.

PR #909, opened 2015-06-10
Added the VS2015 cache directory to the gitignore

rroman81:
Small modification to the gitignore file to ignore the ".vs/**" contents from VS2015.


jansenbe:
Thanks Roman. I'll merge this pr as it helps prepare for VS2015. However keep in mind that we're not yet switching to 2015 and as such we prefer VS2013 solutions. Once we make the switch we'll communicate that via the regular channels like yammer and the monthly call

rroman81:
No problem. I will continue using the VS2013 for any new work, but will have vs2015 for my client's team, so we can be on standby for testing etc.

PR #919, opened 2015-06-11
check if field is assoc to a web or the list

braegelno5:
changed in ObjectListInstance UpdateField


braegelno5:
I will close this now, and will come up with another PR, where we have less queries for the same field.. To check if field exist and update it

PR #929, opened 2015-06-13
fixed field update in list provision, if field is assoc to list - we …

braegelno5:
…check on list if field exists, but then we get an error if we try to get the field from web, because sometimes a field is associated to the list, it's not a site column


braegelno5:
close because summary patch #953

PR #937, opened 2015-06-15
Fix provisioning engine for CSOMv15 support + custom action token replacement

ivanvagunin:
1) Fixed content types provisioning crash on-premises
2) Provisioning engine should not replace URL tokens in custom action ScriptSrc. When ScriptSrc set to absolute URL causes errors like this and appears as a blank page: 
Microsoft.SharePoint.SPException: Cannot make a cache safe URL for "/siteassets/main.debug.js", file not found. Please verify that the file exists under the layouts directory.


erwinvanhunen:
Thank you Ivan for the contribution! Much appreciated.

I will partly merge your solution. The reason is that while you're correct about tokens like ~sitecollection and ~site should not be parsed by the engine but rather by the server, removing the ToParsedString() from the method will also disable the use of any other tokens too. We will implement an overload to the ToParsedString method that takes an optional array of tokens to skip, and we will make sure that ~site and ~sitecollection is not being parsed in the context of the custom action.

ivanvagunin:
Thanks Erwin! That's really a better solution, I tried to minimize impact but could find any better than removing ToParsedString completely. I would not dare to modify TokenHelper on my own )

PR #943, opened 2015-06-16
fix for update list on-prem

ivanvagunin:
CSOM v15 does not support MajorWithMinorVersionsLimit so UpdateList crashes on-premises


erwinvanhunen:
Thank you Ivan! Good catch! Will merge.

Issue #906, opened 2015-06-09
Site Fields which are Lookup Fields are created before the List which is to be used in the lookup have not been created yet. 


NPrice99:
When a site field Is created and the list that is to be used for the lookup has not yet been created, results in an invalid field where the ListName is blank.(see attached file)
eg When I create a lookup field using the provisioning engine, this is the XML which is generated :- <Field Type="Lookup" DisplayName="Directorate" Required="FALSE" EnforceUniqueValues="FALSE" List="Lists/Directorate" WebId="59eea740-3e15-49b8-942e-78f890024f6f" ShowField="Title" UnlimitedLengthInDocumentLibrary="FALSE" Group="My Columns" ID="{3e39d7cd-9d83-42de-b15c-a1f54610e91a}" SourceID="{59eea740-3e15-49b8-942e-78f890024f6f}" StaticName="Directorate" Name="Directorate" />  NB the List is Lists/Directorate.  The attached file shows the resultant field.  Should there not be an error message generated when this happens ?

![lookupfieldsevidence](https://cloud.githubusercontent.com/assets/8484048/8057890/2f3dde6c-0ead-11e5-8aa1-1729a24cfe39.png)

braegelno5:
at the moment all lists are saved to the template not in order on dependencies to each other
we should implement this
we should check this in ProvisioningTemplate and also in ProvisionObjects

PaoloPia:
We do support lookup fields and we provision lookup fields after lists, so that we can find the source list. Are you still experiencing issues with this? Have you seen this sample template?

https://github.com/OfficeDev/PnP-Provisioning-Schema/blob/master/Samples/LookupField.xml

NPrice99:
Hi Paolo The issue is you need to define fields before you can define Lists, and this is the order the provisioning engine does things in.  However, there are lookup fields which need lists to be available before the field can be created and this is the problem.  We almost need the provisioning engine to do two passes at the template.  One to create all of the fields which are not reliant on a list being there and a second pass when all of the lists are there, to create all of the lookup fields which are dependant on there being a list being there so that list can be used to look things up in.

braegelno5:
I have run an debug test against this problem, nprice99 makes the point clear, that's why I wrote we must implement this, it's not yet implemented but in my head..

braegelno5:
So the update for loockup field was implemented as an object event, but it is not working yet, maybe there we need also an update for the web I'd.. Digging

[...]

PR #944, opened 2015-06-16
fix for #906, now we update also the WebId to field definition

braegelno5:


braegelno5:
close because summary patch #953

PR #951, opened 2015-06-16
Fix to update list with more than one contenttype

afsandeberg:
When updating list and adding more than one contenttype en exception will occur.


erwinvanhunen:
Thank you for your contribution Mårten! I will merge.

PR #953, opened 2015-06-17
patch for #906, all updates on lookup fields are now complete handled…

braegelno5:
… so we have now properly working lookup fields if template provisioning is finished
patch replaces #944 and #929 and fixes the issue #906

### DIFF
--- a/OfficeDevPnP.Core/OfficeDevPnP.Core/AppModelExtensions/FieldAndContentTypeExtensions.cs
+++ b/OfficeDevPnP.Core/OfficeDevPnP.Core/AppModelExtensions/FieldAndContentTypeExtensions.cs
@@ -47,7 +47,7 @@ namespace Microsoft.SharePoint.Client
                 throw new ArgumentNullException("DisplayName");
             }
 
-            FieldCollection fields = web.Fields;
+            var fields = web.Fields;
             web.Context.Load(fields, fc => fc.Include(f => f.Id, f => f.InternalName));
             web.Context.ExecuteQueryRetry();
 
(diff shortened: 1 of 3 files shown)


erwinvanhunen:
Hi,

Thank you so much for your contribution! 

I will check out this pull request. Please be careful when using Resharper refactoring/fixup steps, as it causes a lot of small changes that (typed vars that are changed to var, enumerations that are changed to an array, etc.), as this adds to the amount of checking we have to do before we can merge in a PR.

NPrice99:
Hi Erwin - can you let me know when this is available in dev branch.  I can then test it.

braegelno5:
Hi NPrice99 you are able to go to the PR and show all files (3) by clicking on 'view' button, than copy the content over your open file in your VS and run your tests

erwinvanhunen:
Please wait with testing. I am currently looking at the code and not everything in the PR will make it into the dev branch, e.g. I'm 'cherry picking'. I'm reverting some of the refactoring among others.

braegelno5:
And yes erwinvanhunen, after each refactoring i run the test  for this methods and also my own template scenario;)

erwinvanhunen:
Some of your refactoring updates actually changes the behavior of a method, and those I revert right now for instance.

Take for instance the CreateContentTypeFromXML method. In there you write

 if (xDocument.Root == null) return null;

In the previous version an exception would be thrown, in your method it returns null. That's a big change in behavior. While it might make sense to do what you did in the method, we cannot change the behavior of methods, other than fix bugs as it might affect existing code out there.

braegelno5:
you are absolute right! should i build a patch where only the fixes are in, or have you detected them also, yet?

erwinvanhunen:
I'm on it. No need to do a new PR or update the PR. I'm half way already with it.

erwinvanhunen:
I merged your PR with my own branch and have been making some changes to it. I will merge my own branch with the dev branch after I'm done testing and your PR will be automatically merged then too. 

I assume you've been using ReSharper when looking at some of the refactoring in your PR. While the refactoring that R# proposes are of course valid, they not always add to the readibility of the code in our opinion. Especially if the change proposed means that the flow of the code changes, while retaining exactly the same functionality, it's not adding that much other than that it makes it complex for us to validate the code.

I see you added some 'continue' statements here and there, to simplify / remove nested ifs. While continue statements are absolutely valid to use, I vote to use them only at the beginning of a method, effectly making them a condition for the method to execute. When used somewhere deeper in the method, they could become a bit of a hidden 'trap', and are easy to miss when reading code. I removed a few of them and refactored the code back to if blocks.

In general, if possible, it helps us a lot if you make small changes and create multiple commits instead of one big commit with lots of changes. It makes it easier and faster for us to see what has changed and what the possible impact is.

braegelno5:
I will do my best on the next. Grosses Pionierehrenwort;)

NPrice99:
Thanks Erwin !

NPrice99:
Does this change affect Get-SPProvisioningTemplate as well as Apply-SPProvisioningTemplate ?

erwinvanhunen:
It does not affect those cmdlets immediately. You would have to rebuild the PowerShell solution to pick up this change. Or you alternatively build the core library only and replace the OfficeDevPnP.Core.dll in the installation folder of the powershell cmdlets.

braegelno5:
Hi Erwin, a general question for list handling: the generation of  template pulls all lists from context web to the client and then filter them, so we process only the not hidden lists. 
I had changed this in my pr, so we do this filtering on the server, This was also in lookup handling changed. You reverted this. 
My question is, do we need in the future eventual hidden list for a possible business case? And i will answer this for me with yes, was this a flaw from me? Or..?

erwinvanhunen:
It wasn't a flaw from you. Just a mix up from my end. I'm updating the code soon.

PR #954, opened 2015-06-17
Update TimerJob for O365 governance site enumeration

LiyeXu:
1. Make OfficeDevPnP.Core.Framework.TimerJobsTimerJob.sitesToProcess a
protected property, so that a concrete timer job would have access the
total workload amount.
2. Use GetAppOnlyAuthenticatedContext method to create site enumeration
client context object for SPO.

### DIFF
--- a/OfficeDevPnP.Core/OfficeDevPnP.Core/Framework/TimerJobs/TimerJob.cs
+++ b/OfficeDevPnP.Core/OfficeDevPnP.Core/Framework/TimerJobs/TimerJob.cs
@@ -52,7 +52,7 @@ namespace OfficeDevPnP.Core.Framework.TimerJobs
         private string enumerationDomain;
         // Site scope variables
         private List<string> requestedSites;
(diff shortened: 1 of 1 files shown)


jansenbe:
Hi Eric, will have a look on this PR (probably tomorrow)

jansenbe:
Hi Eric,

I'm not merging these changes cause there are alternative means to do what you need:
*/ If the timerjob needs to run using app-only authentication then you can let configure that via the UseAppOnlyAuthentication(ClientId, ClientSecret) method
*/ To add sites to the timer job you can call the AddSite(site) method...if your Job would need to 'override' the sites it gets passed you can use the UpdateAddedSites and ResolveAddedSites to plug in your custom logic

There's more technical documentation about how the timerjob framework works here: https://github.com/OfficeDev/PnP/blob/master/OfficeDevPnP.Core/TimerJob%20Framework.md. Also feel free to ping me so that we have a chat on this.

PR #964, opened 2015-06-18
Core.TaxonomyPicker Term Labels

sstfnv:
Added support to display term labels in the suggestion dropdown.

image


jansenbe:
Thanks Stefan. Nice addition...I've merged this PR

PR #966, opened 2015-06-19
XInclude support for XMLTemplateProvider

ivanvagunin:
Added basic support for XInclude (http://www.w3.org/TR/xinclude/) in Provisioning engine XMLTemplateProvider. Using XInclude elements common parts of XML templates can be shared across multiple templates (e.g. common fields and content types).


PaoloPia:
Ivan, on behalf of the PnP Core Team I want to thank you for your contribution, we really appreciate your commitment.

PR #981, opened 2015-06-24
fix for provisioning extensibility provider Enabled

ivanvagunin:
Provisioning engine ignores Enabled attribute for extensibility providers


jansenbe:
Nice catch Ivan! I'll merge this one

PR #982, opened 2015-06-25
 Update the access token ExpiresOn comparison logic  

LiyeXu:
Recently we've been experiencing access token refresh failure in AutoSites, which is because of the ExpiresOn property is in UTC time and the comparison is using a local current time. Just added a private helper to calculate the lease time instead of the original comparison logic.


jansenbe:
Thanks for this fix Eric. I'll merge this one

LiyeXu:
Cheers!

PR #989, opened 2015-06-28
Core.cross domain lib

stijnneirinckx:
New sample that shows how you can use javascript on a SharePoint page to call serverside MVC methods of a provider hosted app (in another domain). To make this work, I wrote a small cross domain library. The serverside code can create clientcontext in the same way a normal provider hosted app would do. (user and app permissions). It is also possible to show html rendered by the razor views on a SharePoint page.


jansenbe:
Thanks Stijn. Before we can process this sample you'll need to get the CLA signed (https://cla.microsoft.com).

jansenbe:
Excellent work stijn! Another way to beat iframes :)

PR #992, opened 2015-06-29
Correction - Namespace and issue with folder name

aaclage:
Correction of namespace and onload action to load footer after HTML without MDS activated.


aaclage:
Thank you for the fast approval.

PR #993, opened 2015-06-30
Execute query retry fixes dev

Zimmergren:
Found a few places where ExecuteQueryRetry wasn't used, so I changed it to use that instead of ExecuteQuery. Small and simple fixes.


Zimmergren:
Realized that the initial commit, merge branch master and merge branch dev might be incorrectly added (git magic). Really the commits we want are the "Replaced ExecuteQuery...".

erwinvanhunen:
Thank Tobias!

PR #999, opened 2015-06-30
Azure ad.redis cache user profile information from Azure AAD

levalencia:
This samples shows how to use Azure Redis Cache with ADAL Library to get user profile information from Azure Active Directory in and out from the cache to provide a better performant app.


levalencia:
pull request is wrong with master.

PR #1000, opened 2015-06-30
Azure ad.redis cache

levalencia:
This samples shows how to use Azure Redis Cache with ADAL Library to get user profile information from Azure Active Directory in and out from the cache to provide a better performant app


levalencia:
@jansenbe this is the sample I was talking about, azure redis cache works in multi instance websites

jansenbe:
Thanks Luis. Nice one :)

PR #1001, opened 2015-06-30
Fixed the Microsoft.Online.SharePoint.Client.Tenant dependency in the on-premise nuspec file.

rroman81:
Looks like the v15 version contains local depedency for Microsoft.Online.SharePoint.Client.Tenant.dll. I fixed the location to be used from bin/Release15 location.


jansenbe:
Hi Roman, can you explain more about why you want this fix. These files are stored in PnP for reference, trying to understand why you want to change them. We update our nuget packages on a monthly basis and while doing so we generate a new nuspec file. Do you have an issue with our nuget package?

jansenbe:
Closing this PR since we're maintaining these nuspec files...they're copied on a monthly basis as part of the master merge process.

Issue #1014, opened 2015-07-03
 Calculated Field creation order


braegelno5:
Calculated field in list are not able to create by the engine before all referenced columns from the formula are present

erwinvanhunen:
I personally don't think this is an issue but a fact. I mean, the template just has to provide the fields in the correct order.

braegelno5:
normally yes you are right, but that's why i have done this over the sp ui, and saved the template,
so the other way is to do a reorder on template generation to put all calculated fields on the end of list fields. For this we need pr #1015 not.

PR #1015, opened 2015-07-03
fix for issue #1014

braegelno5:
calculated field creation must be done after all fields referenced in formula
https://www.yammer.com/itpronetwork/#/Threads/show?threadId=555937464


erwinvanhunen:
You have a good reasoning here. However, this would mean that the order of the fields in which they go into the list will be changed. E.g. calculated fields will always show up in the bottom of the list, and as a result, in the bottom of the default list forms. This would affect templates already out there which have the fields in a working order in the template, as calculated fields will show up in the bottom for new deployments. We decided that that is un unwanted side effect of this PR and we will not merge this one. Nevertheless, I do think the reasoning behind this PR is valid, and we need to implement something like this in the future, however we then need to retain the field order as defined in the template too. I will make an issue to put this action on the todo list.

PR #1017, opened 2015-07-03
provision engine - view scope provision support

ivanvagunin:
Added support for setting View.Scope property during provision - currently it is ignored


erwinvanhunen:
Thanks Ivan! Will merge

PR #1032, opened 2015-07-09
Added ADFS authentication option for Connect-SPOnline

GeiloStylo:
Added the possibility to use ADFS authentication when connecting to an on-prem SharePoint farm: 
```
Connect-SPOnline -Url http://yourlocalserver -Credentials (Get-Credential) -AdfsHostName 'sts.consoso.com' -RelyingPartyIdentifier 'urn:sharepoint:contoso'
```


erwinvanhunen:
Thank you for your contribution! We did some testing against an on-prem farm with an internal ADFS endpoint and with an O365 environment and we noticed that we can normal authenticate with the cmdlets as they are. Is there a special case you're addressing with this PR?

GeiloStylo:
Yes, maybe it's a special case. The scenario is as described here: https://github.com/OfficeDev/PnP/blob/master/OfficeDevPnP.Core/SAML%20authentication.md. When I used the cmdlets as they are, I always get a 403 error.

erwinvanhunen:
Thank you for the update. I'll merge with a minor modification (due to an update done to the same cmdlet yesterday).

PR #1034, opened 2015-07-10
Skip token replacement in Field Objecthandler

SchauDK:
When provisioning SiteColumns we need to skip replacement of ~sitecollection token because the field can contain a JSLink with the token.


erwinvanhunen:
Good catch! I will merge this one, and will add some additional changes too immediately after the merge through a new merge.

PR #1050, opened 2015-07-16
The call to "update list" throws an exception for document libraries.

mcnga1:
This error occurs when running the provisioning console against a just-provisioned schema that includes a document library, and the pnp:ListInstance does not include EnableAttachments="false".

https://msdn.microsoft.com/EN-US/library/microsoft.sharepoint.splist.enableattachments.aspx
The EnableAttachments property does not apply to any list that has a base type of Survey or DocumentLibrary.
If you set this property to true for either type of list, it throws an SPException.


erwinvanhunen:
Good find! I will merge this one.

PR #1064, opened 2015-07-22
Fixes issue with not being able to add subnodes to top navigation

erwinvanhunen:


laskewitz:
Thanks for the fix! It works like a charm...

PR #1073, opened 2015-07-27
in debug mode we will delete the existing list with possible orphaned external event receiver

braegelno5:
in debug mode we will delete a existing list, so we prevent our system from orphaned event receicers..
https://www.yammer.com/itpronetwork/#/Threads/show?threadId=563126266


jansenbe:
Thanks Torsten. Useful addition, I've added some more comments inline

PR #1077, opened 2015-07-28
Implemented method CleanupTermStore

nanddeepn:
Cleans up the entire term store entries by term group name.


jansenbe:
Thanks Nanddeep, will be looking into this.

jansenbe:
Hi Nanddeep, we're not going to merge this extension method as there are standard API's to delete the termsets and termgroups. We would rather ensure that the developer in question understands that all termsets will be deleted and with the OOB API's he's forced to code it like that. Hope you can understand.

nanddeepn:
Thanks Bert

PR #1084, opened 2015-07-29
Added support for files in nested folders

lestersconyers:
I've added support for uploading files that are nested in folders.


jansenbe:
Thanks Lester. I'll be looking into this.

PR #1085, opened 2015-07-29
Added incremental and debug activation

lestersconyers:
Added the option to only deploy files that have been changed since the console app last finished running. Console app now accepts "activateIncremental" and "debug" as arguments. Debug allows for F5 running to deploy branding files without copying the files to the bin directory. Just add "debug" "online" to the Start Options.


jansenbe:
Thanks Lester. I'll be looking into this.

PR #1091, opened 2015-07-30
Enable provisioning of additional Groups

russgove:
Hi, 
I'm new to git, so I'm not sure if im doing this properly. I've enabled provisionioning of additional sharepoint groups using the web templates.


russgove:
need to make additional gourps optional

PR #1093, opened 2015-07-31
Microsurvey sample (take 2)

BobGerman:
As shown in the "Transitioning to the App Model" panel at Ignite 2015, here is a web part that can be flexibly provisioned in several ways. The web part presents a "microsurvey" - a single question that users can answer and then see the overall responses in place. It now includes custom New, Display, and Edit list forms as well as the web part and settings page. 

This sample can be provisioned as a SharePoint Hosted Add-in, as a "drag-and-drop" solution (a user drags the files into place manually), or in a centrally managed environment such as an organization with dozens or thousands of project or team sites.

In addition to being useful to end users, the sample demonstrates:

- how to write code that is reusable as a SP Hosted App, drag and drop app, or centrally hosted scripts
 - how to write a solution that provisions its own SharePoint storage in JavaScript
 - how to do remote provisioning with the "desired state" pattern, so the script can be re-run to upgrade when the app is versioned
 - how to provision lists with custom forms that won't break when the out of the box forms change
 - how to write an AngularJS "widget" (web part) that allows more than one copy on a page, and shares the page with other AngularJS code
 - how to use the PnP PowerShell commands to deploy a widget and custom site settings page


jansenbe:
Thanks for this contribution Bob. I'm going to have a look at this one - apologies for the delay in PR processing

jansenbe:
manual merge due to restructuring...sample is in so closing the pr

PR #1126, opened 2015-08-07
Provisioning engine extensions - navigation

ivanvagunin:
I have few extensions for provisioning engine that can may be helpful in different scenarios. Here is the navigation provider implementation (can provision both structural and taxonomy navigation) with sample data. I'll try to prepare and add some more providers later.

Not sure if Scenarios is proper folder, but looks like it is.


erwinvanhunen:
Great work Ivan! I actually wonder if this should be provided as an extension though. It should actually be part of the full engine instead I think. But we'll review and discuss it and we will get back to you!

ivanvagunin:
Would be great to have it as a part of engine! Also I was thinking it might be a good idea to make extensibility providers interface similar to existing object handlers, so it would be easier to integrate extensions code to engine. At least method to extract entities (not only provision) with extension providers could help.

ivanvagunin:
Added security extensibility provider

s-KaiNet:
It will be great, if this become part of the core library. Especially security is extremely needed thing. Great work, thanks.

jansenbe:
PR will require manual porting to new PnP-Sites-Core repo + merging with current code

ivanvagunin:
good. should I close this PR?

jansenbe:
If you want to port this change to the PnP-Sites-Repo and integrate with the current state of the engine then that would be great....and it would mean that this PR can be closed. 

I'll leave this one open until either you do this and close the PR or someone of the core team picks up this PR and manually merges it in PnP-Sites-Core

jansenbe:
Hi Ivan,

Thanks for contributing this PR! 

This PR has been open for a long time due to the fact that we've been restructuring the PnP repositories. Today the provisioning engine updates happen in the PnP-Sites-Core library and a lot of changes have been implemented over the last months. 

Therefore it's very hard for us to merge this PR as is...we would like to ask you to resubmit this PR (if still relevant) as a new PR in the PnP-Sites-Repo. We apologize for the extra efforts this is resulting into.

This also means that we're going to close this PR in the PnP repo.

Best regards,
The PnP Core team

PR #1151, opened 2015-08-28
Filter out invalid web features causing problems.  Add original ID to list instance to fix lookup field creation. Support root site collection as template. Optimize check if field should be added, Title is removed from exclusion list to support rename. Fix for calculated field creation, since formula in SchemaXml is not in a working format to provision. Make async job to be able to run from console by. Set MaxDequeueCount to 1 so no exceptions occure when the site collection is already busy creating.

EagleWizard28:


jansenbe:
Hi Mathieu, thanks for this contribution.

You've created this PR against our master branch whereas we only merge PR against our dev branch. Second complication is that in the meanwhile we've split the big PnP repository in several smaller ones...and the core component that your change is targeting is moved to the PnP-Sites-Core repository. 

If you want to have this change being processed I would recommend apply your changes again but this time in your forked version of the PnP-Sites-Core repo. Once that's done you can issue a PR against the DEV branch of the main PnP-Sites-Core repo.

Hope the instructions are clear, if not feel free to reach out.

Given we can't merge this PR I'm going to close it.

PR #1153, opened 2015-08-30
Support for SharePoint hosted Provisioning Templates

wictorwilen:
Introducing SPSiteTemplateManager as well as support for only using
Azure Web Sites configuration (instead of we/app.config)


wictorwilen:
Update to documentation will come shortly wrt to config of SharePoint Only config as well as the Azure Web Sites only config

frankmarasco:
Thank you Wictor for your contribution.

PR #1154, opened 2015-08-30
Fix typo

pascalberger:
Fix typo in readme


pascalberger:
See #1155 for PR with correct target branch

PR #1211, opened 2015-11-06
jQuery People Picker: Make search work in all browsers

pascalberger:
Sometimes e.key can be undefined, which results in a currentValue containing the string `undefined`. For jQuery events `e.which` should always be set and can be converted using `String.fromCharCode`.


retostadelmann:
+1

pascalberger:
Closed this one since it was against master. Opened a new PR against develop: #1213

PR #1213, opened 2015-11-09
Make search work in all browsers

pascalberger:
Sometimes e.key can be undefined, which results in a currentValue containing the string undefined. For jQuery events e.which should always be set and can be converted using String.fromCharCode.


pascalberger:
@patrick-rodgers Any chance to get this merged?

PR #1231, opened 2015-11-18
Keep-alive for Free tier Azure Web Jobs

wictorwilen:
An option to make sure that free tier Azure Web Jobs are kept alive and
not shut down.

### DIFF
--- a/Solutions/Provisioning.UX.App/Provisioning.Common/Office365SiteProvisioningService.cs
+++ b/Solutions/Provisioning.UX.App/Provisioning.Common/Office365SiteProvisioningService.cs
@@ -122,7 +122,7 @@ namespace Provisioning.Common
                     throw;
                 }
                Log.Info("Provisioning.Common.Office365SiteProvisioningService.CreateSiteCollection", PCResources.SiteCreation_Creation_Successful, siteRequest.Url);
-            }, 25000);
+            });
         }
 
         /// <summary>
--- a/Solutions/Provisioning.UX.App/Provisioning.Job/App.config
+++ b/Solutions/Provisioning.UX.App/Provisioning.Job/App.config
@@ -1,59 +1,65 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
-    <configSections>
-      <sectionGroup name="pnp">
-        <section
-          name="tracing"
-          type="OfficeDevPnP.Core.Diagnostics.LogConfigurationTracingSection, OfficeDevPnP.Core"
-          allowLocation="true"
-          allowDefinition="Everywhere"
-      />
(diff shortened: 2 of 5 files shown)


VesaJuvonen:
Hi Wictor,
valid case and does solve the challenge. I'll run this through with Azure CAT to see if there's any other way to get this "fixed", so that we can have a pattern as default model for web jobs.

wictorwilen:
If you run it with the CAT team they prolly make sure it doesn't work anymore...

VesaJuvonen:
Azure has following noted in their documentation - "Web apps in Free mode can time out after 20 minutes if there are no requests to the scm (deployment) site and the web app's portal is not open in Azure. Requests to the actual site will not reset this.". at https://azure.microsoft.com/en-us/documentation/articles/web-sites-create-web-jobs/

Now the key question is that if we do workaround for this and automate the requests to scm, is this sustainable model for the future or not? - Hunting still down right contact for this, but could be falling into grey area. Stack Overflow shows nicely the demand for this, so trying to get official answer rather than guessing on something which might go away in next weeks.

speschka:
I tried multiple ways…like every way I could think of to do keep alives of free web sites to take advantage of web jobs in free azure sites.  In every case azure was “smart enough” to realize that it was not a “real user” (probably related to the client IP, since all my work arounds also ran in Azure) and so none of them kept the site alive.  After spending considerable time on this I finally abandoned this approach and moved to one of the paid tiers of Azure web sites to one that was always alive.  And then for other reasons ultimately moved it out of web jobs altogether and back to where this code really should run, which is a cloud worker role.

I think it’s popular because most people see it the same way I did – as a possible free alternative to a typically expensive resource, the cloud worker role.

My $.02.

Steve

wictorwilen:
This is interesting. Seems like they did something on the 24th of November to prevent this! Logs shows that everything worked up until then as expected. Need to look into this further...

VesaJuvonen:
Thanks Steve for sharing that and not that surprised on the fact that it would be now blocked. Let's keep this PR open for now if Wictor will test things still, but sounds like this isn't the right solution for now.

VesaJuvonen:
Just to process this, let's close this for now. We can re-open the PR if needed later or submit a new one if there's alternative way to do this.

wictorwilen:
An update: it seems to work now, so we're not being blocked doing this, but there's no guarantee. And the documentation actually says "Web apps in Free mode can time out after 20 minutes", notice "can".

VesaJuvonen:
Since the wording is indeed "can" and we cannot rely on the model, would not get this included in the default sample. It would not be anyway suitable model for any mid-sized or enterprise deployment, even though it would save few bucks from developers.

PR #1262, opened 2015-12-19
prevents an RER to refire itself (if we make update in ItemUpdated, …

braegelno5:
…the ItemUpdating is fired .. | again .. | again..)

### DIFF
--- a/Samples/Core.EventReceivers/Core.EventReceiversWeb/Core.EventReceiversWeb.csproj
+++ b/Samples/Core.EventReceivers/Core.EventReceiversWeb/Core.EventReceiversWeb.csproj
@@ -57,6 +57,7 @@
       <Private>True</Private>
     </Reference>
     <Reference Include="System.IdentityModel" />
+    <Reference Include="System.Runtime.Caching" />
     <Reference Include="System.Runtime.Serialization" />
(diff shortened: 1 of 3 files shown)


jansenbe:
Thanks Torsten...looks interesting, but we did not yet have time to test this. Will be for the next master merge

jansenbe:
Hi Torsten, I've been looking into this implementation. You've indeed raised a problem and found a solution for it. Your solution however will not work reliable in an enterprise context because one typically has load balanced the RER. You could store the cache in a database and then this could work but that would mean a lot of infrasture and work for such a basic requirement.

Wouldn't it be better if you check the before/after properties in the RER and depending on that decide to either update the item or skip the update. Stina Qvarnström has nicely described this pattern here: http://tech.bool.se/how-to-stop-the-itemupdated-event-from-refiring-itself-in-an-remote-event-receiver/.

Given that Stina's approach is light-weight without infrastructure dependencies I think it's better to recommend that one. I'll update the documentation of this sample to spell this out.

Anyhow, thanks for raising the issue for the creative solution!

jansenbe:
Documentation updated. 

Will close this PR now, thanks again for making us aware of this issue.

PR #1274, opened 2016-01-07
Cleanup readme of Core.UserProfiles.Sync sample

pascalberger:
Use tenantid placeholder instead of specific OneBit tenant in readme


jansenbe:
Thanks Pascal

PR #1286, opened 2016-01-18
Fixed possible encoding issues when search term contains special characters

Fensterbank:
The search term was directly included in the URL which results in encoding issues on server side when the request was sent by Internet Explorer and umlauts has been contained in the search term.
By using encodeURIComponent all special characters are encoded before performing the request.


jansenbe:
Thanks Frédéric, I'll have a look at this PR.

jansenbe:
Thanks Frédéric, good change so have merged this one.

PR #1292, opened 2016-01-22
Made a minor change in the GetUsersFromActiveDirectory method. Added …

svarukala:
Made a minor change in the GetUsersFromActiveDirectory method. Added comment on the change within the code. The do-while loop incorrectly misses the last batch of users. The condition I added checks for the last batch of users and adds it to the list. Otherwise, the pagedCollection.MorePagesAvailable fails in the while condition and the loop misses the last batch of users.


jansenbe:
Thanks svarukala. Good change, will merge this one.

PR #1307, opened 2016-02-10
New Sample: MicrosoftGraph.Office365.GroupsExplorer

pschaeflein:


VesaJuvonen:
thx Paul, will have a look early next week.

pschaeflein:
Found some bugs as part of the rename. Looks like the fix is included in this PR. Let me know if I'm interpreting this screen wrong.

VesaJuvonen:
Thanks Paul, that looks good now. Sorry for the small delay on getting this merged. Will be done for the March release anyway.

pschaeflein:
No problem. Let me know if you need anything.


- Paul

VesaJuvonen:
Actually - sorry for immediate follow up. Currently sample is using beta end points cross all the requests. That's absolutely needed for some of them, but any change updating those end points to v1.0 which are available. Will increase the life cycle of the sample after we can lock the end point to right version. Makes sense?

pschaeflein:
Thanks for the reminder. I ran into issues just today during a training session where the beta endpoint returns something different than when I first wrote it.

I’ll switch the endpoints in the next few days.


- Paul

pschaeflein:
@VesaJuvonen - changed endpoints

VesaJuvonen:
Thx Paul, will get this merged for March release. Would you be available to do a 10-15 min demo on this during the March community call on 8th of March at 8 AM PST?

pschaeflein:
Yes, I would be delighted to run thru the code. Thanks!

VesaJuvonen:
thx Paul. Great to have this included and let's absolutely get short demo recorded in monthly community call.

PR #1325, opened 2016-02-29
dist support using gulp, browserify, tsify. gulpfile.js refactoring also

darei-fr:
Added dist support in gulp of the typescript files using tsify & browserify.
pnp.core.js
pnp.core.min.js
pnp.core.min.js.map

Also provided some global variables to the gulp file to ease the understanding.
Quick question of what's the value of require if we use browserify (inside a single file ?)

Still working on the SPDevExperience (adding files into SharePoint) & watchify for changes into SP also.

### DIFF
--- a/Libraries/Core.JavaScript/bower.json
+++ b/Libraries/Core.JavaScript/bower.json
@@ -1,0 +1,24 @@
+{
+  "name": "PnP.Core",
+  "description": "A reusable library",
+  "main": "PnP.Core.js",
+  "authors": [
+    "Microsoft and other contributors"
+  ],
+  "license": "MIT",
+  "homepage": "https://github.com/OfficeDev/PnP/",
+  "moduleType": [
+    "amd",
+    "globals",
+    "node",
+    "yui"
+  ],
+  "ignore": [
+    "**/.*",
+    "node_modules",
+    "bower_components",
+    "test",
+    "tests"
+  ],
+  "private": true
+}
--- a/Libraries/Core.JavaScript/gulpfile.js
+++ b/Libraries/Core.JavaScript/gulpfile.js
@@ -3,7 +3,9 @@
 //******************************************************************************
 //* DEPENDENCIES
 //******************************************************************************
+
 var gulp = require("gulp"),
+    print = require('gulp-print'),
     browserify = require("browserify"),
     src = require("vinyl-source-stream"),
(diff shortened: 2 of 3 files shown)


darei-fr:
First version of bower support also.

darei-fr:
Added standalone support on browserify, seems working with UMD bundle & server-root.

olemp:
@darei-fr Cannot find module 'merge-stream'. Is this supposed to be installed globally?

darei-fr:
Yeah, it seems we have a issue with that, and it was fixed by adding it to package.json by some folks. But I wonder if we really need this import today. I don't think we use it.

olemp:
Yeah, I just added to my local package.json.

So we might just remove it from dependencies? It's not used in the gulpfile.

darei-fr:
Yes good idea, I have added the PR.

darei-fr:
@olemp you mean without browserify etc ? 
If you have other ideas, it's more than welcome.

darei-fr:
Normally the gulp package command produce a bundle of all ts script with browserify (pnp.min.js, etc).
If i use "gulp serve-dist" minified files are copied into server-root/scripts. Require pnp works in the html file, i see the random string.

Having said that, i know that there is some discussion to split the full module into multiple one. I didn't have this concept in mind, there will be again some changes after our meeting i think :)

Actually, I was trying to bundle all the files to a single one to ease the upload to a SharePoint library.

darei-fr:
Based on feedbacks, i can change the implemention of bundles.

What i like :
 - One file, easy dist & sp sync eventually
 - Dependencies handled by Browserify
 - On the fly compilation of ts files & live reload

What i dislike : 
  - To produce the bundle, browserify use tsify, i would prefer to use tsc.

So if you have another better way to minify, i'am on it :)

olemp:
I think the 'package' tasks is fine, but I can't get the output dist to work. Probably not related to this PR?

When reffing the dist .min.js or .js I should be able to call the getrandomstring function, but pnp is not defined for me.

darei-fr:
I will fetch to try to reproduce

olemp:
@darei-fr This might be worth a look https://github.com/abogartz/typescript-amd-gulp-example/blob/master/README.md

patrick-rodgers:
Closing this PR - I manually merged it and will ensure you get the credit for the work. Thanks!

Issue #1327, opened 2016-03-01
Core.JavaScript: Implement sharepoint/data


olemp:
Ease use of REST APIs.

Functions for e.g.
* Fields
* Content Types
* Lists 
* List Views
* List Data
* Web Properties
* Page Properties

olemp:
Would prefer to use Acccept-header application/json;odata=nometadata, but due to onprem support, we should stick to application/json;odata=verbose.

patrick-rodgers:
Perhaps that can be configurable. I would also really like to see us aim for a fluent API structure, so something like pnp.sharepoint.rest.site.web.lists.getbytitle("title").items.select("Title", "Summary").filter({}).get() <-- which returs a promise. So using the TypeScript interfaces as a way to provide intellisense support for all the various calls. I'd start out small, like getting site, web, and list data. Work on the pattern before going too deep. Look forward to what you can come up with!

olemp:
I think I really like that style @patrick-rodgers. 

When it comes to executing these calls, should we add a dependency on jQuery (or similiar), or just use stick to vanilla XMLHttpRequest?

olemp:
Will start work on this here (https://github.com/olemp/PnP/tree/ISSUE-1327) @PaoloPia

olemp:
Intial structure: https://github.com/olemp/PnP/commit/50da39bf7e5b7e1d719186778a772a4f4a0d4165
Web and Site classes: https://github.com/olemp/PnP/commit/50da39bf7e5b7e1d719186778a772a4f4a0d4165

patrick-rodgers:
We talked about that a little in the meeting. I think it makes sense to go ahead and add a dependency to jQuery, but we can discuss more in the meeting. The new SharePoint UI fabric uses it, and so does everyone else. Plus their ajax is solid and very well documented with tons of available examples. We do want to wrap the returned Deferred object in an es6 style promise, likely using the library es6-promise, see thread [here](https://www.yammer.com/itpronetwork/#/Threads/show?threadId=665618813). Perhaps reach out to Arto to coordinate.

olemp:
Ok.

My initial thought if we're skipping jQuery was to create ajax.ts under /utils using XMLHttpRequest. It would have a get() and post() method and they would both return a ES6 Promise.

We're using the ES6 Promise polyfill yeah?

Any input @artokai?

patrick-rodgers:
[...]

PR #1338, opened 2016-03-02
Implemented sharepoint/rest base structure (Issue #1327)

olemp:
So far the following are supported:

.....web.get()
.....web.lists.get()
.....web.lists.getByTitle().get()
.....web.lists.getById().get()

All returns an ES6 promise (but no polyfill included yet).

and select() and filter()  are always available from the Queryable base class.

I haven't included jQuery in the dist or anything yet, so it's a dependency at the moment (need to be referenced before requiring pnp).


olemp:
Added initial tests for Queryable, Web and Lists.

The tests are making sure the correct URLs are returned by Queryable url() function.

In my opinion there's no point in having mock data for the fluent API when we're building on top of the REST API. How would you create mock data for every single endpoint we're supporting? It might be possible, but I don't think it's necessary. 

The important thing here is that the calls yields the correct URLs.

**...web.get()**

Does url() return 
```javascript
_api/web? 
```
Yes? Test OK.

**.. web.lists.get()**

Does url() return 
```javascript
_api/web/lists?
```
 Yes? Test OK

**... web.lists.select(["Title"]).get()**

Does url() return 
```javascript
_api/web/lists?$select=Title
```
Yes? Test OK

patrick-rodgers:
I agree, the tests should ensure we are getting the urls we expect from the builder. We can do other testing with data later, but for now we are most concerned with are we building the correct url.

Issue #1328, opened 2016-03-01
Port over configuration data


patrick-rodgers:
As discussed in yammer conversation:

https://www.yammer.com/itpronetwork/#/Threads/show?threadId=665618813

PR #1347, opened 2016-03-07
JSSIG - Added configuration objects to the global PnP namespace

artokai:
Related to issue #1328: Port over configuration data

Added configuration and related providers to the global PnP object graph in order to have them included also in the packaged and minimized (``gulp package``) output. Also fixed some import paths that were not working during the packaging step.


patrick-rodgers:
Thanks Arto - nice work. Possible I messed up doing the merges this weekend so nice catch on the path updates.

PR #1392, opened 2016-04-06
Update CustomInjectedJS.js

anibakore:
There is a small bug in the original code. When user clicks on the back button, the event handler doesn't fire and the URL is not replaced. By simply adding _spBodyOnLoadFunctionNames, the function is called after the body is loaded again and the link is replaced correctly every time.

Change:
_spBodyOnLoadFunctionNames.push('SubSiteOverride_OverrideLinkToAppUrl');


patrick-rodgers:
Are you sure this change is correct? In the modified file you are pushing the result of the function call into the onload functions array. Would it not be better as you have it in your comment above to push the function name?

anibakore:
Yes. There was a typo in the modified file. Corrected the file and creating a new pull request.

patrick-rodgers:
Thanks!

Issue #1426, opened 2016-05-17
People picker issue with email address containing apostrophe


dlwDenysma:
PnP/Components/Core.PeoplePicker/Core.PeoplePickerWeb/Scripts/peoplepickercontrol.js
#### Category

[X] Bug
#### Environment

[X] Office 365 / SharePoint Online

We recently had an issue at a customer where an email address contained an apostrophe. According to the spec RFC 2822 (https://tools.ietf.org/html/rfc2822) This is actually possible. Up until now I was not aware of this. 
Without this fix, we could not select a user with an apostrophe in the email address from the people picker, change this:

`txtResults += this.Format(displayTemplate, this.InstanceName, loginName.replace("\\", "\\\\"), this.HtmlEncode(displayName), email, displayName, loginNameDisplay, title);
`

into this:
`
txtResults += this.Format(displayTemplate, this.InstanceName, loginName.replace("\\", "\\\\"), this.HtmlEncode(displayName), this.HtmlEncode(email), displayName, loginNameDisplay, title);`

Dean144:
While the standard may allow apostrophes, MS does not allow them in the mail attribute, see https://support.office.com/en-us/article/Prepare-to-provision-users-through-directory-synchronization-to-Office-365-01920974-9e6f-4331-a370-13aea4e82b3e?ui=en-US&rs=en-US&ad=US,

PR #1427, opened 2016-05-17
emails containing apostrophe fix

dlwDenysma:
Fix for issue #1426


jansenbe:
Hi dlwDenysma,

Thanks for this contribution. It looks absolutely OK, but unfortunately it's targeting our master branch whereas we can only accept PR's against our DEV branch. Would it be possible to resubmit this PR against our DEV branch?

Thx!

